PR #2, opened 2026-09-03
Harden the untrusted-input paths

prognost1c:
## Harden the untrusted-input paths

A security review of the archive, convert and preview paths turned up one
confirmed command-execution bug from a crafted filename, plus several places
where an untrusted decoder could run outside the sandbox. Each fix is small,
covered by the existing suite, and adds no dependency (`Cargo.lock` is
unchanged; the tree stays zero-crate).

### fix: terminate option parsing before untrusted operands
`compress_argv`'s tar branch handed the selected file names to `bsdtar`
straight after `-C <parent>` with no `--`. A file literally named
`--use-compress-program=<cmd>` in a compressed directory was parsed by bsdtar
as an option and ran `<cmd>`. `trash.rs` already inserts `--`; compress now
does too. `convert_one` canonicalizes its input to an absolute path so
ImageMagick cannot read it as an operator, and an archive `op` that is neither
`compress` nor `extract` is refused by name instead of falling through to
extract. Added a unit test proving the `--` sits before a dash-leading member.

### fix: require the sandbox for archive, convert and probe children
`run_boxed`, the archive index reader, the preview lister and the ffprobe
probe ran unsandboxed when `sandbox::available()` was false. Thumbnails already
fail closed there; these now do too. `util-linux` (ships `prlimit`) is added to
`depends` so the jail's second half is guaranteed present alongside
`bubblewrap`.

### fix: copy a share link without a shell
The Dropbox share URL was interpolated into `bash -c "... | wl-copy"`;
`$(...)` and backticks in the URL would be expanded. Now `["wl-copy", url]`.

### fix: open the copy source with O_NOFOLLOW
Closes a same-uid TOCTOU where a source swapped to a symlink after the lstat in
`copy_any` would be followed by `copy_file`.

### Testing
`cargo test --release --locked` → 336 passed, 0 failed. `tests/js.sh` → 1013
checks, 0 failed. `tests/keymap-gen.sh` → ok. Verified end to end against the
built backend: a bogus op is refused by name, and with the jail forced
unavailable the archive path refuses the job rather than running the tool
unsandboxed.

### Deliberately left as follow-ups
- The read loop's stdin lines are uncapped (same-uid client only) — needs a
  reader rework, filed separately.
- `sandbox.rs` ro-binds all of `/etc`; trimming it risks breaking decoders that
  read fonts/mime and the network is already unshared.
- `is_runnable` uses `metadata` (follows symlinks) so a `/usr/bin` symlinked
  thumbnailer still resolves; switching to `symlink_metadata` would reject
  legitimate ones. The Exec runs sandboxed with no network regardless.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved archive handling for filenames beginning with hyphens.
  * Prevented unsafe symlink following when copying files.
  * Prevented shell interpretation when copying share links.
  * Validated archive operations and rejected unsupported requests.
  * Converter and media probing now receive safer, canonicalized paths.
* **Security**
  * Archive, media, and metadata operations now fail safely when sandboxing is unavailable instead of running without protection.
* **Reliability**
  * Improved handling of archive and thumbnail processing in restricted environments.

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -7,7 +7,8 @@ pkgdesc='Fast, keyboard-first file manager for Omarchy'
 arch=('x86_64')
 license=('MIT')
 # omarchy owns /usr/share/omarchy/shell, which ui/Commons and ui/Ui link into; quickshell owns qs.
-depends=('bubblewrap' 'glib2' 'omarchy' 'quickshell' 'shared-mime-info' 'xdg-utils')
+# util-linux ships prlimit, which the thumbnail and archive sandboxes require alongside bubblewrap.
+depends=('bubblewrap' 'glib2' 'omarchy' 'quickshell' 'shared-mime-info' 'util-linux' 'xdg-utils')
 makedepends=('cargo')
 optdepends=('libarchive: archive listing and extraction'
             '7zip: 7z archive support'
--- a/src/backend/archive.rs
+++ b/src/backend/archive.rs
@@ -73,6 +73,10 @@ impl Formats {
         a.push(dest.to_string_lossy().to_string());
         a.push("-C".to_string());
         a.push(parent.to_string_lossy().to_string());
+        // Members are relative names from the selection; one that begins with a dash would be read
+        // as a bsdtar option (`--use-compress-program` runs an arbitrary program), so option parsing
+        // is terminated first, the same way trash.rs does before its own paths.
+        a.push("--".to_string());
         a.extend(names.iter().cloned());
         Some(a)
     }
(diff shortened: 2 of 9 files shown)


thisisgm:
Merged in `bfb9b31` (`v0.1.3`): the `--` terminator, the three fail-closed sandbox paths, the argv-direct `wl-copy` and `O_NOFOLLOW` all landed, plus adversarial cases for the three refusals and a sandbox refusal that now names the tool it would have run.

PR #4, opened 2026-09-03
fix: drop the box-shadow under overlays

smillunchick:
Floating menus and dialogs no longer sit on a drop shadow. The dimmed
backdrop behind dialogs is unchanged.

./tests/js.sh: 997 checks, 0 failed.
tools/flea-qmllint-gate: 0 regressions.

## Summary by CodeRabbit

* **Visual Changes**
  * Removed drop shadows from context menus, conversion dialogs, network dialogs, and keymap sheets.
  * Floating surfaces now render without the previous shadow layers, resulting in a flatter appearance.
* **Removed**
  * Removed configurable shadow styling from the application theme.
  * Removed the shadow rendering component and related shadow configuration options.


thisisgm:
Merged in `8d86230` (`v0.1.3`): overlays are flat, and the canvas is being corrected to match so the board and the app agree.

PR #5, opened 2026-09-03
feat: the rail pans, one modified wheel map over both surfaces, and the arrows step between the views

joewinke:
## What this fixes

Two operator-reported gaps, both in the same neighbourhood:

1. **The rail could not scroll.** The rail was a bare `Column` under its surface rectangle with no viewport at all, so a wheel over it did nothing, and a rail that outgrew its own height (many favourites, several mounts, a NAS and a Dropbox row) could not show its bottom rows by any means. The list's wheel verb is a pan, and the rail's is now the same one: the rows live inside a `Flickable`, the plain wheel pans, and nothing chases the cursor back — the same split of the two verbs `List.qml` draws.

2. **The arrows were dead keys while browsing.** `Left`/`Right` were bound to `seekBack`/`seekForward`, which resolve to nothing until a preview is open, so the rail was reachable only through Tab. With no preview open and no grid, `Left` in the list now steps into the rail and either arrow in the rail steps back out; Tab and Esc keep exactly what they had.

## The modified wheel, one map for both surfaces

A shared `ui/js/Wheel.js` routes the chords, because the same modifier has to mean the same thing over both surfaces or the split between them is a bug neither file can see:

| gesture | list | rail |
|---|---|---|
| plain wheel | viewport pans, cursor follows (unchanged) | viewport pans (new — this was the bug) |
| `Alt`+wheel | step the cursor a row a notch | step the rail cursor |
| `Ctrl`+wheel | jump to first/last row | jump to first/last rail row |
| `Shift`+wheel | extend the selection (`Shift+Up/Down`'s verb) | not accepted — rail rows carry no selection, so it pans |

Two environment facts the map honours, both verified by driving events and reading what the app actually received:

- the reporting operator's input stack **rewrites Alt+wheel onto the horizontal axis** (`angleDelta.x = ±120`, `angleDelta.y = 0`, modifiers intact) — caught by a probe that logged the raw event. `Wheel.axisDelta()` reads whichever axis moved, so the rewritten chord and the ordinary one answer the same verbs, on this box and everywhere else.
- a natural-scroll setup inverts the axis before the event arrives, so `Wheel.direction()` folds in the event's `inverted` flag and both surfaces step with the hand, not the raw axis.

The plain wheel is never accepted by the handler, so it falls through to the Flickable. On the list side the chords ride `Filter.moveCursor` / `Filter.setCursorView`, which already scroll the row they land on into view; on the rail side every cursor move — keyboard, chords, the clamp — reveals its row through the Flickable.

## Why a MouseArea and not a WheelHandler

The first attempt used `WheelHandler` with `acceptedModifiers`, and the chords never fired: a Flickable consumes every wheel its children sit under before any handler in the tree can answer one. Driven on a live desktop with injected wheel events carrying real modifier state, the handlers stayed silent over both the ListView and the rail while the view under them scrolled. The chords now arrive through a `MouseArea` overlay (`acceptedButtons: Qt.NoButton`, so presses and drags are untouched) which declines the plain wheel and answers the chords.

## What moved, for the budget

`ui/Sidebar.qml` was at its cap, so the change carries two cuts the budget's own story prescribes:

- the favourites merge (Home first, first position wins) moved whole to `Places.favorites`
- the rail menu's chosen-row dispatch (key, not position — the five-second-poll problem) moved to `Mounts.release`

## Tests

- `tests/js/wheel.js` checks the modifier map, the clamps, the axis fold, the direction fold, and the end each wheel direction lands on, wired into the harness beside the rest
- `tests/js/focus.js` carries the three new lookups beside the two it dropped, which had asserted the old silence
- `./tests/js.sh`: **1035 checks, 0 failed**; `tools/flea-file-budget` exits 0 (soft warnings unchanged from upstream)

Both surfaces were also driven against the installed QML on a live Hyprland desktop: the rail pans under a rail taller than the window, the plain wheel pans the list with the cursor following, Shift extends in both directions, and the arrows walk into and out of the rail.

One deliberate non-change: the grid keeps `Left`/`Right` for stepping tiles and both previews keep them for seeking and paging, so nothing that already meant something changed meaning.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -3596,3 +3596,41 @@ key press fires a compositor bind, so `omarchy-drive hotkey --global super w` is
 on the ACTIVE window, so assert the active window is the one the run launched before sending it, and
 never call `hl.dsp.window.close()` with no argument: it returns `ok` rather than printing a signature
 and acts on whatever is active, which may be a window you did not open.
+
+### The rail pans, the arrows step between the views, and the wheel chords share one map
+
+Three operator reports drove one branch: the rail could not scroll at all, the rail that outgrew
+its own height could not show its bottom rows by any means, and the arrows were dead keys while
+browsing, because `Left`/`Right` are bound to `seekBack`/`seekForward`, which resolve to nothing
+until a preview is open.
+
+**The rail's rows live inside a `Flickable` now.** The rail was a bare `Column`, and a wheel over
+it did nothing: every row is already drawn (see `railItemFor`'s no-virtualization note), so a
+scroll had no viewport to pan and no verb to answer. The list's wheel verb is a pan, and the
+rail's is the same one now, with `boundsBehavior: StopAtBounds` and a clipped viewport; nothing
+chases the cursor back, which is the same split the list draws.
+
+**A Flickable consumes every wheel its children sit under before any WheelHandler in the tree can
+answer one.** The first chord attempt used `WheelHandler` with `acceptedModifiers`, and over both
+the ListView and the rail the chords were silent while the view under them scrolled — driven on a
+live desktop with injected wheel events carrying known modifier state, handlers attached as far up
+as the window root and as deep as the row under the pointer. The chords therefore arrive through a
+`MouseArea` (`acceptedButtons: Qt.NoButton`, so presses and drags are untouched): a plain wheel is
+declined and the view pans natively, the modified chords are answered through `ui/js/Wheel.js`.
+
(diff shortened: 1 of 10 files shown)


thisisgm:
Merged the rail viewport in `9a4e6e3` (`v0.1.3`); the wheel chord map and the arrow refocus are not in, because the swapped-modifier reading is a synthetic-injection artefact rather than a property of the box, and `revealCursor` needed both `mapToItem` arguments before the viewport worked at all.

PR #8, opened 2026-09-03
feat: add directory tabs in the window

smillunchick:
A Flea window can hold more than one directory. `t` opens a tab at the current folder, `w` closes it, and `1` through `9` switch.

The strip appears only after a second tab, so a single listing keeps the chrome the first paint uses. Closing the last tab is refused. Nine is the cap, because those digits are the jump keys.

One listing stays live. A hidden tab is a snapshot (path, cursor, view, history, sort, selection), not a second view. That keeps per-file work inside the viewport.

`:` is still not built and still says so.

## Test plan

- `./tests/js.sh` — 1036 checks, 0 failed
- `./tests/ui.sh tabs` — not run here (`omarchy-drive` was missing)
- On a live window: `t`, navigate, `1`/`2`, `w`, last-tab refuse

## Summary by CodeRabbit

- **New Features**
  - Added directory tabs with support for opening, switching, and closing tabs.
  - Added keyboard shortcuts: `t` to open, `w` to close, and `1`–`9` to switch tabs.
  - Added a tab bar with selectable tabs, close controls, and a new-tab button.
  - Tab state preserves folder location, cursor, view settings, sorting, and selections.
  - The tab bar appears when two or more tabs are open, with a maximum of nine tabs.
  - Prevented closing the final remaining tab.

- **Tests**
  - Added comprehensive tab behavior and UI test coverage.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -445,6 +445,7 @@ huge pages" below for what it is worth and what it cost.
   `flea --open`, see "Opening a file".
 - `ui/ContextMenu.qml` is the one pane-owned right-click popup and its single Open action.
 - `ui/StatusBar.qml` renders the path, row counts and transient messages.
+- `ui/TabBar.qml` is the window's tab strip, hidden with no height until a second tab exists.
 - `keys.toml` is the one key table, and `tools/flea-keymap-gen` turns it into `Keymap.js`.
 - `ui/js/Keymap.js` is the generated key-to-action lookup and imports no QML.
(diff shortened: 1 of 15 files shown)


thisisgm:
Merged in `b867075` (`v0.1.3`): tabs ship and `tests/ui.sh tabs` passes on the box; a tab opened from search results now records where you were rather than the scope the search walked, and no switch that re-lists carries a selection, because indices name different rows once the listing is re-read.

PR #10, opened 2026-09-03
feat: the header's own right-click menu, column show/hide behind it, and an Advanced group

joewinke:
Follows #5 (independent of it — branched from main).

## What this adds

**A right click over the column titles** opens the pane's one ContextMenu — the same single instance the listing and the rail raise, because a second instance in this tree takes the whole window's keyboard (see AGENTS.md) — with:

- one toggle row per optional column: **Hide/Show Mode, Size, Date Modified, Kind**. Name is not offered: it is the one column a file manager cannot do without, and `ui/js/Columns.js` already refuses to drop it.
- an **Advanced ▸** flyout holding the state toggles: **Show/Hide hidden files** (moved here from its top-level row, so the next preference is one row in the group and not another top-level row the menu never had room for).

Left click on a title still sorts; `sortable` still gates that.

## How the pieces fit

- **Columns**: `Columns.set(width, tokens, hidden)` subtracts the user's set from what the width affords — width still wins, so "Show Kind" on a 600 px pane stays dropped rather than crowding the name. The set lives in a new `ViewState` singleton (qs module, beside Theme) backed by a FileView at `~/.config/flea/view.json`, the same read/watch/write pattern the bookmarks use, so the choice outlives the window.
- **The menu construction moved to `ui/js/Menu.js`** (`listingEntries`, `headerEntries`, `advancedRows`), so menu rows are unit-testable without a window — and the move paid ContextMenu's headroom back with interest: it is 21 lines lighter carrying a third face.
- **The Open row now names the application** xdg-open would choose — "Open" beside the resolved name in muted type (`ui/MenuRow.qml`'s new suffix slot) — resolved as the cursor moves by `ui/Opener.qml` (the one foreign-program owner), empty for a directory, which flea opens itself. **Copy Path** sits beside Open and hands the row's full path to `wl-copy` through the same owner; flea's own copy clipboard (`Ops.clip`) is a different thing and stays one.

## Tests

- `tests/js/menu.js`: the listing construction (Open's suffix, Copy Path's place, the Advanced group, the empty-listing shape) and the header construction (flip labels, `col:<key>` verbs, Advanced carrying the toggle)
- `tests/js/columns.js`: the hidden set subtracted from the width-afforded set, width winning, and the no-set callers unchanged
- `./tests/js.sh`: **1026 checks, 0 failed**; `tools/flea-file-budget` exits 0 (Pane lands exactly at its 400 cap, ContextMenu 21 lines lighter)

## Summary by CodeRabbit

* **New Features**
  * Right-click file-list headers to show, hide, and manage columns.
  * Hidden-column preferences are saved and restored between sessions.
  * Context menus include Copy Path and display the default application for files.
  * Header menus provide column visibility, hidden-file, and folders-first settings.
  * Folders-first sorting can be toggled and persisted across sessions.
  * Move to Trash now offers an in-menu confirmation step.
* **Bug Fixes**
  * Column layouts consistently respect hidden columns and available width.
  * Context-menu actions, submenu navigation, and empty or directory listings behave more consistently.
  * Menus prevent unintended pointer interactions with underlying rows.

### DIFF
--- a/docs/protocol.md
+++ b/docs/protocol.md
@@ -51,7 +51,12 @@ default to `0`.
 
 ### sort
 
-`{"c":"sort","by":"<string>","desc":<bool>}`
+`{"c":"sort","by":"<string>","desc":<bool>,"dirs":<bool>}`
+
+`dirs` is optional and sticky: absent, the session keeps its standing choice (true, folders
+grouped ahead of files, the shipped default); present, it sets that choice for this sort and
+every listing and sort after it, `list` included. `false` interleaves folders into the order
+the key and direction produce.
 
 Example: `{"c":"sort","by":"size","desc":true}`
 
--- a/src/backend/metasort.rs
+++ b/src/backend/metasort.rs
@@ -10,7 +10,7 @@ use std::time::Instant;
 // The stats live for this call alone: nothing is cached, so reversing a size order stats the
 // directory again at the warm cost docs/protocol.md "sort" records, and a listing sitting in name
 // order, which is every listing the field measures, never carries 16 bytes a row it is not using.
-pub fn sort_by_stat(l: &mut Listing, base: &Path, by: SortBy, desc: bool) -> (f64, f64) {
+pub fn sort_by_stat(l: &mut Listing, base: &Path, by: SortBy, desc: bool, dirs_first: bool) -> (f64, f64) {
     let (stats, pass_ms) = stat_all(base, l);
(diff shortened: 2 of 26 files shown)


thisisgm:
Merged the header menu and column show/hide in `e46ea3e` (`v0.1.3`); the Open row application name is out because resolving it per cursor move spawned three processes a keystroke, and the hidden toggle stays a top-level row.

PR #16, opened 2026-09-03
feat: accelerate pointer scrolling across vertical views

stappmus:
## What changes

- adds one shared `FastScrollHandler` with a 4x multiplier
- applies it to the list, grid, Miller columns, places rail, text preview, and zoomed PDF pane
- handles smooth `pixelDelta` input and discrete `angleDelta` wheel notches
- clamps every write to the Flickable origin and content end

## Why 4x

This ports the behavior from [Omarchy Spotify commit 7499123](https://github.com/stappmus/Omarchy-Spotify/commit/74991237230ba29afa1300cdd895b7272f45dbd5), where the multiplier was calibrated to feel about as fast as Chromium.

Discrete wheels still honor `Application.styleHints.wheelScrollLines`; the multiplier sits on top of that platform preference. Touchpads keep their fractional pixel deltas, multiplied by the same amount.

## Event handling

The handler uses a `MouseArea` overlay with `acceptedButtons: Qt.NoButton`, so taps and drags remain owned by the existing row handlers. This follows the live finding from #5 that a child `WheelHandler` loses wheel events to its `Flickable` parent.

The scroll stays immediate and animation-free, matching the existing motion rule. A wheel event is accepted only when it actually moves the target, so boundary events can continue propagating.

## Verification

- `cargo build --release`
- `./tools/flea-qmllint-gate` — 0 actionable regressions
- `./tools/flea-file-budget` — passes; existing soft warnings unchanged
- live Wayland smoke launch reaches `Configuration Loaded`

No new automated tests are included.

## Summary by CodeRabbit

* **New Features**
  * Added faster scrolling to lists, grids, sidebars, PDF pages, and text previews.
  * Improved mouse wheel and touchpad scrolling with configurable speed and bounded movement.
  * Preserved normal tapping and dragging interactions while enhancing wheel-based navigation.

### DIFF
--- a/ui/ColumnPane.qml
+++ b/ui/ColumnPane.qml
@@ -46,6 +46,11 @@ Item {
         boundsBehavior: Flickable.StopAtBounds
         reuseItems: true
 
+        Flea.FastScrollHandler {
+            parent: view
+            flickable: view
+        }
+
         delegate: Flea.ColumnRow {
             required property int index
             width: view.width
--- a/ui/FastScrollHandler.qml
+++ b/ui/FastScrollHandler.qml
@@ -1,0 +1,53 @@
+import QtQuick
+
+// Multiplies both smooth touchpad deltas and discrete wheel notches, then writes the bounded
+// position directly. A MouseArea is intentional: a Flickable consumes wheel events before a child
+// WheelHandler can answer them, while acceptedButtons: Qt.NoButton leaves taps and drags alone.
+MouseArea {
+    id: root
+
+    required property var flickable
+    property real speedMultiplier: 4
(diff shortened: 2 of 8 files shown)


nixfred:
I reproduced the same scrolling issue and found two related overflow gaps: long SMB share lists were not scrollable, and large Taildrop/context menus could extend offscreen. I opened stappmus/flea#1 against this PR with those fixes plus deterministic wheel/bounds and nine-surface coverage: https://github.com/stappmus/flea/pull/1

thisisgm:
Merged in 5063929 (v0.1.6) with your commit as its parent: one notch is four discrete steps, touchpad pixels move one to one on both axes; thanks.

PR #21, opened 2026-09-03
Network places: edit, remove, and live SFTP matching

TomFaulkner:
A network place can be added, and `r` can relabel it, but there was no way to change the URI or drop the bookmark. On a fresh Omarchy install that showed up immediately: SFTP bookmarks were saved, one used a literal `~` (`gio` does not expand it), and fixing the line meant editing `~/.config/gtk-3.0/bookmarks` by hand.

This PR is the rail work that followed from that, not only the edit dialog.

## Edit and remove

- Right-click (or `m`) on a NETWORK share offers **Edit** and **Remove**.
- An unmounted bookmark is still reachable, so a URI that will not mount can be rewritten.
- A mounted share keeps **Unmount** first, then Edit, then Remove.
- Edit opens the existing add-location form, filled from `Protocols.parse`.
- Save rewrites that bookmark line in place via `Places.replace` (same GTK file, same normalized match `relabel` already uses).
- Remove drops the matching bookmark line. If the share is live, it also unmounts.
- Dropbox install stays on the add dialog only.
- `r` is unchanged: it still only relabels.
- Ctrl+E cannot fire Edit. It reads `Mounts.releaseAction` (eject/unmount only) instead of `railMenu()[0]`.

## Live mounts vs bookmarks (SFTP)

The add form writes `sftp://user@host:22/path`. `gio mount -l` reports the same share without the default port, so opening a saved place used to add a second, brighter row and leave the cursor on the dim bookmark. `normalize` now drops a scheme's default port on write and on compare.

Once collapsed, the live `gio` row kept gio's own name (`tom` for an SFTP session) and dropped the bookmark's, so rename wrote the file and the next poll put `tom` back. `Places.networkEntries` keeps the bookmark label on the live row.

gvfsd-sftp mounts **one connection per host**. `gio` lists `sftp://user@host/` even when the bookmark is `…/home/tom`, so a second path on that host stayed dim, and Unmount of the root failed as busy while Flea was listing `/home/tom` on the same FUSE mount. Extra SFTP/FTP paths on a live host now stay their own rows and read as mounted. Unmount sends gio the live root URI and goes Home first if any path on that connection is open. SMB/NFS stay per-share.

## Tests

- `tests/js.sh`: parse round-trips, replace/remove, rail menu, Ctrl+E, live/bookmark merge, SFTP connection coverage, SMB still per-share.
- `tests/ui.sh case_networkedit` seeds a `~` bookmark, opens Edit, rewrites the path, and asserts the bookmarks file.
- `case_unmount` expects Unmount, Edit, Remove on a live share; Return still fires Unmount.

`tests/ui.sh` was not run on the original box (`omarchy-drive` is not installed). JS suites and the file-budget hard cap passed.

## Notes

`gio` does not expand `~` in an SFTP path; use `/home/user` (or leave Path empty for the remote `/`). Hyprland also does not start `gvfsd-fuse`; without it Flea can mount over D-Bus and then fail with "no browsable folder". That daemon is a session concern, not part of this PR.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -3621,3 +3621,18 @@ key press fires a compositor bind, so `omarchy-drive hotkey --global super w` is
 on the ACTIVE window, so assert the active window is the one the run launched before sending it, and
 never call `hl.dsp.window.close()` with no argument: it returns `ok` rather than printing a signature
 and acts on whatever is active, which may be a window you did not open.
+
+### Editing a network place rewrites the bookmark line, not just its label
+
+`r` on a NETWORK row still only relabels. The URI itself is edited through the same dialog Add
+uses: right-click (or `m`) on a share, **Edit**, which fills `ui/NetworkForm.qml` from
+`Protocols.parse` and Save writes through `Places.replace`. An unmounted bookmark's menu is Edit
+alone, so a URI that will not mount (the operator's real `sftp://user@host:22/~`; gio does not
+expand `~`) can still be rewritten. A mounted share keeps **Unmount** first so Ctrl+E still
+releases; `Mounts.releaseAction` is what Ctrl+E reads, never `railMenu()[0]`, because Edit is a
+menu row and not a release.
+
+`Places.replace(body, oldUri, newUri, label)` is the write: normalized match like `relabel`, every
+duplicate rewritten, control characters stripped from the label, an empty old URI appends (Add).
+`tests/js/places.js`, `tests/js/protocols.js`, `tests/js/mounts.js` and `tests/ui.sh case_networkedit`
+hold it.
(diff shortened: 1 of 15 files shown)


TomFaulkner:
Addressed the two review comments on the latest commit:

- `Protocols.parse` now splits off the path before looking for `@`, so a path like `inbox@2026` round-trips.
- Remove of a live share unmounts first and rewrites the bookmarks file only if `gio mount -u` succeeds. An idle bookmark is still dropped immediately. `tests/ui.sh case_networkremove` covers the failed-unmount case.

thisisgm:
Partly shipped in `9c13aeb`, released in `0.1.4`: the rail menu gains Rename and Remove, and a scheme default port no longer splits one share into two rows; the Edit dialog that rewrites a bookmark URI and the per-connection SFTP handling are not in it, so this stays open for those.

thisisgm:
Landed in v0.3.0 reworked onto the dialogs board with credit; closing in favour of that commit, thank you.

PR #31, opened 2026-09-03
feat: identify remote-to-remote transfers

nixfred:
## What changes

- detects when every source and the destination are under distinct GVFS mount roots
- labels those operations as `Copying between remote hosts` or `Moving between remote hosts`
- keeps mixed local/remote batches and same-mount operations under generic progress wording
- keeps the existing backend, cancellation, partial-file cleanup, and undo behavior unchanged

## Why

Remote-to-remote transfers already use Flea's normal backend through their GVFS paths, but the progress line presented them as ordinary local work. The explicit label makes the network path and its likely latency clear without adding a second transfer implementation.

## Review fixes

- require distinct remote roots before applying the remote-to-remote label
- clear pending transfer classification when a transfer is refused before it starts
- cover both copy and move wording
- fail the shell integration test if the move never completes

## Verification

- [x] `./tests/run-all.sh` — 10 suites, 0 failed
- [x] `./tests/js.sh` — 1,173 checks, 0 failed
- [x] `./tests/ops.sh` — synthetic SFTP-to-SMB copy and move, including source retention/removal
- [x] mixed local/remote and same-mount batches are not mislabeled
- [x] existing failure, no-clobber, cancellation, and undo tests remain green
- [x] `./tools/flea-qmllint-gate` — 0 regressions
- [x] `./tools/flea-file-budget`
- [x] `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added support for identifying transfers between two remote GVFS mounts as remote-to-remote operations.
  * Remote-to-remote copy and move progress now shows the transfer direction, item count, and filename.
  * Transfers between files on the same remote mount are recognized separately.

* **Bug Fixes**
  * Prevented failed transfers from leaving stale transfer status that could affect subsequent operations.
  * Improved handling of mixed local and remote file selections.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ Large directories stay responsive: the window loads rows and requests thumbnails
 - **Three views.** List, columns and grid, with tabs and natural filename sorting.
 - **Quick Look.** Press Space for images, PDFs, text, media and archive contents.
 - **File operations.** Copy, move, rename, trash, compress and extract, with an undo journal.
+  Copies and moves between two network mounts run on that same undoable backend and are identified
+  as remote-to-remote.
 - **Network and sharing.** SMB, SFTP, FTPS, WebDAV, NFS, Dropbox and Taildrop.
 - **Desktop integration.** Default file manager, “Show in folder” and Open/Save dialogs.
 - **Your settings.** Omarchy text sizes, configurable menus and Default, Vim, Mac or Windows keys.
--- a/tests/js/harness.qml
+++ b/tests/js/harness.qml
@@ -28,6 +28,7 @@ import "places.js" as PlacesSuite
 import "protocols.js" as ProtocolsSuite
 import "railkeys.js" as RailKeysSuite
 import "recent.js" as RecentSuite
+import "remote.js" as RemoteSuite
 import "renderer.js" as RendererSuite
 import "scroll.js" as ScrollSuite
 import "search.js" as SearchSuite
@@ -70,7 +71,7 @@ Item {
             ["palette", PaletteSuite], ["pathbar", PathBarSuite], ["places", PlacesSuite],
             ["picker", PickerSuite],
             ["protocols", ProtocolsSuite], ["railkeys", RailKeysSuite],
-            ["recent", RecentSuite],
+            ["recent", RecentSuite], ["remote", RemoteSuite],
             ["renderer", RendererSuite],
             ["scroll", ScrollSuite], ["search", SearchSuite],
             ["selection", SelectionSuite], ["settings", SettingsSuite],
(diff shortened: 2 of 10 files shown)


nixfred:
Rebased onto current `main` — this had gone stale behind 479 commits and was conflicting. Thanks for your patience on it.

Three things needed adapting rather than just replaying, so flagging them rather than letting them pass silently:

- **`tests/js/harness.qml`** — the harness moved to a `suites` array since I opened this. The remote suite is now registered as `["remote", RemoteSuite]` alongside the others instead of the old `RemoteSuite.run(check)` call.
- **`README.md`** — the section I originally edited has been rewritten. The sentence now hangs off the current **File operations** bullet instead, saying the same thing in your wording.
- **`ui/PaneWire.qml`** — a genuine overlap: `main` now clears `pane.pendingMenu` on a failed listing, and this branch clears a refused transfer's classification at the same point. They do different jobs, so I kept both rather than picking one.

`tools/flea-qmllint-gate` reports 0 regressions. `tools/flea-file-budget` exits clean; it flags `ui/PaneWire.qml` as over the soft 250-line budget, but it was already 373 lines on `main` before these 4 lines.

I couldn't run `tests/ops.sh` — it needs a `cargo build` I don't have set up here — so that one is worth running on your side.

nixfred:
Good catch from the bot, and it was a real one — pushed in `e958a6d`.

**`Transfer.sampled()` dropping `kind` was the whole feature failing in the ordinary case.** The classification survived `Ops.started` and was gone by the first progress update. `sampled` and `itemDone` rebuild the transfer field by field rather than copying it, so a field they do not name is dropped, and `kind` was not named. The wire calls `sampled` on every byte sample, so a transfer that started as remote-to-remote reverted to "Copying 1 of 2" the instant bytes began arriving — which is the only time the label is on screen long enough to read.

Both carry `kind` now, and the comment above them says why every field has to be restated, so the next field added does not go the same way.

I verified the new tests actually catch it rather than just passing: reverting the one-field fix produces exactly three failures, all of them the generic wording.

```
FAIL a sample keeps the transfer remote-to-remote: got "Copying 1 of 2, photo.heic"
FAIL and so does the item line that ends it: got "Copying 1 of 2, photo.heic"
FAIL a sampled remote move keeps its verb too: got "Moving 1 of 2, photo.heic"
```

**On the `moving: true` coverage** — that one is already there, as "a remote-to-remote move uses the matching verb". I added a sampled version of it anyway, since that is the path that was actually broken, plus a local transfer to prove the plain wording still works.

59 checks in ops, 1852 across `tests/js.sh`, 0 failed. `tools/flea-qmllint-gate` reports 0 regressions.

Still could not run `tests/ops.sh` here — it needs a `cargo build` I do not have set up — so that one is worth a run on your side.

PR #33, opened 2026-09-03
fix: clear output between reusable processes

nixfred:
## What was wrong

The QML services reuse their `Process` objects and keep a parallel `_...Output` value because `onExited` can race `StdioCollector.text`. Those fallback values were never cleared before the next command.

A quiet second command could therefore consume the first command's bytes:

- an empty mount poll retained disconnected shares
- empty `gio info` could reopen the previous share's local path
- an empty server listing could repeat another server's shares
- a mount failure with no stderr could inherit an earlier `already mounted` message
- empty `lsblk` output could retain removed devices
- empty Tailscale status could retain stale Taildrop targets

## Fix

Clear each fallback after rejecting an overlapping run and immediately before starting its reusable `Process`. Timeouts still retain the last good model intentionally; this only prevents a completed new run from borrowing an older run's raw output.

## Verification

- [x] new `tests/process-output.sh` poisons all six fallback values and starts the real QML Process objects
- [x] mutation check: removing one reset makes the new suite fail
- [x] `./tests/run-all.sh` repeated twice — 11 suites, 0 failed; 1,153 JS/QML checks
- [x] `./tools/flea-qmllint-gate` — 0 regressions
- [x] `./tools/flea-file-budget` — all hard caps pass
- [x] `git diff --check`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -3578,6 +3578,17 @@ reachable while a device is misbehaving. The `_streamPending` interlock is relea
 rather than in the timer, because a collector whose stream was cut off may never finish it and
 `poll()` refuses on that flag alone.
 
+**Every reusable Process starts with an empty fallback.** `StdioCollector.text` and the parallel
+`_...Output` value exist because `Process.onExited` can race the collector property update, but the
+fallback is ordinary component state and survives the next command. Before this rule, a quiet
+second run could reuse the first run's bytes: an empty `gio mount -l` retained a disconnected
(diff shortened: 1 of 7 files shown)


thisisgm:
Declined PR #33: Quickshell v0.3.1 replaces each waitForEnd collector body before Process.exited, the public two-run control returned empty without stale data, and no code landed; thanks @nixfred for the report and proposed patch.

PR #37, opened 2026-09-04
fix: build and run on aarch64, and two defects found doing it

### LINKED ISSUE #13: Does not build on aarch64: statfs/renameat2 externs use *const i8, PKGBUILD is x86_64 only
## Environment

- Omarchy 4.0.2 on Asahi Linux (Apple Silicon, `uname -m` = `aarch64`)
- Quickshell 0.3.1, rustc 1.98.0 (Arch `rust 1:1.98.0-1`)
- flea v0.1.2, commit `0654cf1`

## Symptom

`makepkg -si` refuses outright because `PKGBUILD` declares `arch=('x86_64')`. With `--ignorearch`, `cargo build --release --locked` fails:

```
error[E0308]: mismatched types
  --> src/backend/fsinfo.rs:68:24
   |
68 |     if unsafe { statfs(c.as_ptr(), &mut buf) } != 0 {
   |                 ------ ^^^^^^^^^^ expected `*const i8`, found `*const u8`

error[E0308]: arguments to this function are incorrect
  --> src/backend/ops.rs:35:23
   |
35 |     let rc = unsafe { renameat2(AT_FDCWD, c_from.as_ptr(), AT_FDCWD, c_to.as_ptr(), RENAME_NOREPLACE) };
   |                       ^^^^^^^^^
   = note: expected raw pointer `*const i8`
              found raw pointer `*const u8`
```

## Cause

`std::ffi::c_char` is `i8` on x86_64 Linux but `u8` on aarch64 Linux, and `CString::as_ptr()` returns `*const c_char`. The two `extern "C"` declarations hardcode `*const i8`, so they only type-check on x86_64.

## Fix

Two lines, plus the arch list in `PKGBUILD`:

```diff
--- a/src/backend/fsinfo.rs
+++ b/src/backend/fsinfo.rs
@@ -55 +55 @@
-    fn statfs(path: *const i8, buf: *mut StatFs) -> i32;
+    fn statfs(path: *const std::ffi::c_char, buf: *mut StatFs) -> i32;
--- a/src/backend/ops.rs
+++ b/src/backend/ops.rs
@@ -20 +20 @@
-    fn renameat2(olddirfd: i32, oldpath: *const i8, newdirfd: i32, newpath: *const i8, flags: u32) -> i32;
+    fn renameat2(olddirfd: i32, oldpath: *const std::ffi::c_char, newdirfd: i32, newpath: *const std::ffi::c_char, flags: u32) -> i32;
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -7 +7 @@
-arch=('x86_64')
+arch=('x86_64' 'aarch64')
```

The hand-declared `struct StatFs` in `fsinfo.rs` needs no change: glibc's generic 64-bit `struct statfs` has the same field order and widths on aarch64 as on x86_64, so `f_type`, `f_bsize` and `f_bavail` land where the struct expects them.

## Verified on this box after the patch

- `cargo build --release --locked`: OK
- `cargo test --release --locked`: 335 passed, 0 failed
- `flea --backend` fed `{"c":"list","path":"<46-entry dir>","first":5,"hidden":false}` answers `listed` (read 0.028 s) followed by `rows`

I have not installed the package or opened the Quickshell window yet, so this says nothing about the UI or the Vulkan path on Asahi. Can send the above as a PR if that is easier.

the2dl:
Closes #13, and carries two defects that issue does not cover — one of them a security boundary that is open in v0.1.3 today, and one that is not architecture specific at all.

Tested on an M2 Air (`Mac14,15`) running Omarchy 4.0.2 on Asahi, aarch64, 16 KiB pages, glibc 2.43, rustc 1.98.0, quickshell 0.3.1.

## The three fixes

**1. `c_char`, not `i8` — the build stops here (#13).**

`c_char` is `i8` on x86-64 and `u8` on aarch64, and `CString::as_ptr()` returns `*const c_char`, so the hardcoded `*const i8` in the `statfs` and `renameat2` declarations only type-check on one architecture. Same fix @stgomoyaa proposed.

The hand-declared `StatFs` needs no change, and I checked that rather than assuming it: `struct statfs` is byte-identical on both, 120 bytes with `f_type` at 0, `f_bsize` at 8, `f_bavail` at 32, read out of the header on this box with `offsetof`. Only its comment moves.

**2. `O_NOFOLLOW` was the x86-64 value, so the guard did nothing on arm64.**

This one is not in #13, and it is not cosmetic. `copyfile.rs` declared `O_NOFOLLOW` as `0o400000`. arm64 carries its own `asm/fcntl.h`, where `O_NOFOLLOW` is `0o100000` and `0o400000` is `O_LARGEFILE` — a no-op on 64-bit. So `open` accepted the flag and ignored it, and the same-uid race the function documents was open: a source swapped to a symlink after `copy_any`'s stat was followed, and the target's bytes were copied.

The suite already catches it. `a_source_swapped_to_a_symlink_after_the_stat_is_refused_rather_than_followed` fails on aarch64 before this commit and passes after it. #13 did not hit it because `O_NOFOLLOW` landed in b4b7ee4, after the v0.1.2 that issue was filed against — so this shipped in v0.1.3.

Because a wrong number here disables a security boundary in silence rather than erroring, an architecture with no value listed is now a `compile_error!` instead of quietly inheriting one.

**3. The heap guard has never tested anything in a release build — on any architecture.**

`drop(vec![0u8; RATCHET_BYTES])` is dead, so a release build elides the allocation and the free with it. Nothing is mmapped, nothing is freed, the threshold never ratchets, and the assert passes whether or not the pin is live. `check()` builds `--release`.

Measured rather than reasoned about, by making `pin_mmap_threshold()` conditional on an env var:

| | pin live | pin removed |
|---|---|---|
| before this PR | passes 3/3 | **passes 3/3** |
| after this PR | passes 3/3 | **fails 3/3** |

`black_box` is what holds the allocation down. This is the fix I am least sure you will want as written — if you would rather assert on `mallinfo2().hblkhd` than on `/proc/self/maps`, say so and I will redo it.

Separately, the fixed 1 MiB probe is smaller than the arena libtest leaves behind on a 16 KiB page kernel — about 1.9 MiB here — so glibc served it from a free chunk without consulting the threshold at all, and the test failed spuriously in 2 of 3 full-suite runs. The probe is now sized past the live arena, so only the threshold can decide where it lands. In C on this box, with the ratchet forced, a 1 MiB probe lands in `[heap]` even pinned, while a 3 MiB probe lands outside pinned and inside unpinned — which is the discrimination the test wants.

## Tests

On this box, at this branch:

- `cargo test --release --locked` — **338 passed, 0 failed** (6 consecutive full-suite runs, no flakes)
- `./tests/js.sh` — **1153 checks, 0 failed**
- `./tests/keymap-gen.sh` — ok, both checks
- `./tests/protocol.sh` against the installed `/usr/bin/flea` — **110 of 110**
- `makepkg -f` end to end, then `pacman -U`, then `flea --default`: `hyprctl configerrors` clean, both binds resolve to Flea and nothing is left bound to Nautilus

The GUI runs, which #13 explicitly did not claim: `flea --gui` opens the Quickshell window in ~0.25 s on Asahi, and the Vulkan path in rule 5 gives no trouble. Thumbnails work — a 512×512 JPEG through `glycin-thumbnailer` inside the bwrap sandbox in 7 ms.

## Two notes that are not defects

`tests/protocol.sh:268` copies `$FIXTURE_ROOT/flea-media-btrfs/photo_0.jpg` with no fallback, unlike line 255 which has one. Without your media fixture, `photo.jpg` stays the 1-byte `x` from that earlier fallback, no thumbnailer can read it, and three thumbnail assertions fail for a reason that has nothing to do with the code. Dropping any real JPEG at that path turns all three green. I have left it alone since it is your fixture layout, but a fallback on 268 would make the suite self-contained.

The AUR packages are `x86_64` only, so on ARM the repository's own PKGBUILD is the only route. Worth a line in `docs/install.md` if you take this.

https://claude.ai/code/session_01UN8dj8X3y5nQmBg1cgebXH

## Summary by CodeRabbit

* **New Features**
  * Added package support for 64-bit ARM systems.

* **Bug Fixes**
  * Improved filesystem compatibility across supported Linux architectures.
  * Preserved symlink protection on ARM systems and reported unsupported architectures explicitly.
  * Improved heap probing reliability by adapting checks to the live heap size and preventing optimization from skipping test allocations.

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -4,7 +4,7 @@ pkgname=flea
 pkgver=0.1.3
 pkgrel=1
 pkgdesc='Fast, keyboard-first file manager for Omarchy'
-arch=('x86_64')
+arch=('x86_64' 'aarch64')
 license=('MIT')
 # omarchy owns /usr/share/omarchy/shell, which ui/Commons and ui/Ui link into; quickshell owns qs.
 # util-linux ships prlimit, which the thumbnail and archive sandboxes require alongside bubblewrap.
--- a/src/backend/copyfile.rs
+++ b/src/backend/copyfile.rs
@@ -9,8 +9,17 @@ use std::sync::atomic::{AtomicBool, Ordering};
 const CHUNK: usize = 256 * 1024;
 // rename(2) sets EXDEV when the two paths are on different filesystems, which is the one failure that means "copy instead".
 const EXDEV: i32 = 18;
-// open(2) O_NOFOLLOW on linux x86-64; declared here because this tree takes no libc crate.
+// open(2) O_NOFOLLOW; declared here because this tree takes no libc crate. The value is not the same
+// on every linux architecture: arm64 carries its own asm/fcntl.h, where 0o400000 is O_LARGEFILE and
+// therefore a no-op on 64-bit, so a single hardcoded constant drops the guard below without failing.
+#[cfg(target_arch = "x86_64")]
 const O_NOFOLLOW: i32 = 0o400000;
+#[cfg(target_arch = "aarch64")]
+const O_NOFOLLOW: i32 = 0o100000;
+// A new architecture must add its own value rather than inherit one silently: this flag is a
+// security boundary, and the wrong number disables it quietly instead of erroring.
+#[cfg(not(any(target_arch = "x86_64", target_arch = "aarch64")))]
+compile_error!("O_NOFOLLOW is architecture specific: add this target's value from its asm/fcntl.h");
 
 // What a copy reports as it runs; a directory has no total without a sweep, so it reports 0 and renders indeterminate.
 pub struct Progress<'a> {
--- a/src/backend/fsinfo.rs
+++ b/src/backend/fsinfo.rs
@@ -10,7 +10,7 @@ pub fn dev_of(path: &Path) -> u64 {
     use std::os::unix::fs::MetadataExt;
     std::fs::metadata(path).map(|m| m.dev()).unwrap_or(0)
 }
-use std::ffi::CString;
+use std::ffi::{c_char, CString};
 use std::path::Path;
 
 // The f_type values for the filesystems this box can actually mount; anything else reports its hex.
@@ -33,7 +33,7 @@ const MAGIC: &[(i64, &str)] = &[
     (0x1CD1, "devpts"),
 ];
 
-// struct statfs on linux x86-64: the fields this needs are f_type, f_bsize and f_bavail, and the
+// struct statfs on 64-bit linux, identical on x86-64 and arm64 (120 bytes, same offsets): the fields this needs are f_type, f_bsize and f_bavail, and the
 // rest is padding this never reads. Sizes are from man 2 statfs.
 #[repr(C)]
 struct StatFs {
@@ -52,7 +52,7 @@ struct StatFs {
 }
 
 extern "C" {
-    fn statfs(path: *const i8, buf: *mut StatFs) -> i32;
+    fn statfs(path: *const c_char, buf: *mut StatFs) -> i32;
 }
 
 pub struct Info {
--- a/src/backend/ops.rs
+++ b/src/backend/ops.rs
@@ -2,7 +2,7 @@
 use crate::backend::copyfile::{copy_any, Progress};
 use crate::backend::undo::Step;
 use crate::error::{from_io, FleaError};
-use std::ffi::CString;
+use std::ffi::{c_char, CString};
 use std::path::{Path, PathBuf};
 use std::sync::atomic::AtomicBool;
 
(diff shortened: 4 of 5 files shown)


patrikduksin:
+1 on this pr, wanna use flea on my m1pro machine <3

thisisgm:
Shipped in `fe9e114` as #47, released in `0.1.4`: the `c_char` fix, your arm64 `O_NOFOLLOW` finding (the guard was a no-op there, and `src/uistore.rs` carried the same defect, swept in `7c96b08`) and the release-build heap guard all landed.

PR #39, opened 2026-09-04
fix(backend): bound sandbox memory with cgroups

### LINKED ISSUE #17: sandbox: 1 GiB address space limit (--as) exhausts virtual memory in multi-threaded thumbnailers (e.g. glycin-thumbnailer)
### Summary
In `src/backend/sandbox.rs`, `wrap()` and `wrap_readonly()` impose a 1 GiB virtual address space limit on sandboxed thumbnailers:
```rust
const ADDRESS_SPACE_BYTES: u64 = 1024 * 1024 * 1024;
```
which is passed to `prlimit --as=1073741824`.

On 64-bit Linux, `RLIMIT_AS` bounds **virtual address space**, not resident physical memory (RSS). Modern multi-threaded thumbnailers like GNOME's `glycin-thumbnailer` (used for standard JPEG/PNG/WebP/HEIF/JXL thumbnailing under `/usr/share/thumbnailers/glycin-image-rs.thumbnailer`) easily exceed 1 GiB of virtual address space when decoding high-resolution images (such as a 12 MP phone photo).

When virtual address space is exhausted:
1. `mmap` calls for new thread stacks and memory allocations fail with `ENOMEM`.
2. Glibc's `pthread_create` returns `EAGAIN` (`Resource temporarily unavailable`).
3. Rust's worker thread spawning panics:
   ```text
   thread 'blocking-2' panicked at library/std/src/thread/scoped.rs: failed to spawn thread:
   Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }
   ```
4. Simultaneously, LCMS profile allocation fails (`ObjectCreationError`).
5. An unwinding panic reaches an `extern "C"` FFI boundary in `libglycin` (`gly_image_get_specific_frame`), which calls `abort()` / `SIGABRT` (`panic in a function that cannot unwind`).
6. `flea` treats the non-zero exit as a decoding failure on the file and writes a permanent `fail/` marker in `~/.cache/thumbnails/fail/flea/`.

### Steps to Reproduce
1. Have `glycin-loaders` and `glycin` installed (default thumbnailer on modern GNOME/Arch setups).
2. Open a folder in `flea` containing a high-resolution image (e.g., 4080×3072 JPEG, ~5 MB).
3. `glycin-thumbnailer` aborts with `SIGABRT` (exit code 134) and dumps core.

Minimal standalone reproduction:
```bash
prlimit --as=1073741824 bwrap --unshare-all --die-with-parent --new-session --clearenv \
  --ro-bind /usr /usr \
  --ro-bind /etc /etc \
  --symlink usr/lib /lib \
  --symlink usr/lib /lib64 \
  --symlink usr/bin /bin \
  --symlink usr/bin /sbin \
  --proc /proc \
  --dev /dev \
  --tmpfs /tmp \
  --ro-bind <high_res_photo.jpg> <high_res_photo.jpg> \
  --bind /tmp /tmp \
  /usr/bin/glycin-thumbnailer --input file://<high_res_photo.jpg> --output /tmp/test.png --size 256
```
Fails with `failed to spawn thread: Os { code: 11, ... }` and `SIGABRT`.
Increasing `--as` to 2 GiB or omitting `--as` succeeds without error.

### Proposed Solution
In `src/backend/sandbox.rs`:
- Either increase `ADDRESS_SPACE_BYTES` significantly (e.g., 4–8 GiB) to allow for 64-bit multi-threaded runtime VM allocations (glibc malloc arenas reserve 64 MB per arena, thread stacks reserve 8 MB each),
- Or replace `RLIMIT_AS` with a mechanism that bounds actual physical memory rather than virtual address space (or omit `--as` while retaining `--cpu` and bubblewrap sandboxing).

> Filed by Gemini 3.8 Flash via Antigravity.

Yiin:
## Summary

- Replace the 1 GiB address-space limit with cgroup v2 memory limits.
- Keep one delegated scope per thumbnail worker and one disposable cgroup per file.
- Use transient scopes for lower-volume archive and media helpers.
- Keep setup failures separate from decoder failures, so setup failures write no cache marker.

## Why

Glycin can reserve more than 1 GiB of virtual memory for a valid image. RLIMIT_AS kills that decode even when resident memory stays low.

The new limit controls resident memory and swap for the complete helper process tree. It preserves sparse virtual mappings.

## Validation

- `cargo test --release`: 363 passed
- `./tests/run-all.sh`: 11 suites passed
- The ICC-tagged 6000 by 3375 JPEG fails on the old code and passes with this change.
- Resource tests cover memory limits, descendants, cleanup, scope reuse, setup failure, and final exec failure.
- Two final review lanes approved the implementation.

Closes #17

## Summary by CodeRabbit

- **New Features**
  - Added stronger resource controls for thumbnail generation, helping prevent runaway memory use and process leaks.
  - Added safer archive listing with time limits and cleanup of child processes.
  - Added support for ICC-tagged images during thumbnail generation.

- **Bug Fixes**
  - Improved handling of sandbox startup, execution, timeout, and memory failures.
  - Prevented infrastructure failures from incorrectly creating thumbnail failure markers.

- **Installation**
  - `systemd` is now required for sandboxed operations.


thisisgm:
Partly shipped in `8c60318`, released in `0.1.4`: your ICC fixture is what reproduces #17 on the real thumbnail path, but the cap stayed `RLIMIT_AS` at 2 GiB rather than moving to cgroups, so this stays open for that.

PR #42, opened 2026-09-04
fix: aarch64 support for statfs and renameat2 FFI calls

avillagran:
Fix compilation on aarch64 by casting CString::as_ptr() to *const i8 for statfs and renameat2 FFI calls. CString::as_ptr() returns *const u8 on aarch64, but the C function signatures expect *const i8.

## Summary by CodeRabbit

* **Refactor**
  * Updated low-level system call argument handling for improved compatibility with platform interfaces.
  * No user-visible behavior changes.


thisisgm:
Shipped in `fe9e114` as #47, released in `0.1.4`: the `c_char` cast for `statfs` and `renameat2` is what fixed the aarch64 build.

PR #48, opened 2026-09-04
fix(nav): parent reselects the directory you left

shawnyeager:
## Summary

`h` and Backspace still go to the parent. The parent listing now selects the folder you left, so `h` then `l` is a round trip.

`l` as browse-forward is already on main from #25. This PR no longer touches that bind.

`pendingSelect` already re-reveals a row after a refresh. `Nav.parent` now plants it before the climb. Root and an in-flight listing still refuse, and they plant no select.

`./tests/js.sh` after the change: 1178 checks, 0 failed.

### DIFF
--- a/tests/js/nav.js
+++ b/tests/js/nav.js
@@ -17,6 +17,7 @@ function pane() {
         thumbState: "stale",
         dirSizeState: "stale",
         cursorIndex: 7,
+        pendingSelect: "",
         renamingIndex: 4,
         trashArmedAt: 12345,
         listingState: "ready",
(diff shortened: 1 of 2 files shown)


TomFaulkner:
Duplicate of merged PR #25 ?  @coderabbitai

shawnyeager:
@TomFaulkner yes on the `l` bind — that is #25. This PR no longer touches it.

What is left is the piece #25 did not do: `Nav.parent` plants `pendingSelect` on the folder you leave, so `h` then `l` is a round trip.

jessekalil:
One possible follow-up improvement would be to preserve the navigation context for each parent directory, rather than only reselecting the immediate directory after `h`.

For example:

    wallpapers -> images -> Downloads

After pressing `h` twice, the user should be back in `Downloads`. Pressing `l` twice should then retrace the path back to `wallpapers`, with each directory remembering the child that was previously selected.

This is also how Yazi behaves: `h`/`l` handle hierarchical navigation, while `H`/`L` remain a separate back/forward history. Yazi preserves each directory's cursor state, so returning through parent directories with `l` restores the previous path naturally.

The current `pendingSelect` approach solves the immediate `h` -> `l` round trip, but a per-directory cursor/context history could extend it to this multi-level behavior without changing the meaning of `H` and `L`.

This is only a follow-up suggestion and should not block the current PR, which already fixes the immediate parent reselection issue.

thisisgm:
Merged for v0.3.0, thank you.

PR #51, opened 2026-09-05
feat(grid): fit thumbs and marks to the tile

Jijjy:
## Why

Icons and images look too small by default in grid view. The mark is a ~48px slot inside a much wider cell, so photos sit in padding and folder glyphs read as list-row leftovers.

## What

- The thumb/mark slot is the cell width, square, with `Theme.spacing.gap` inset.
- Photos `PreserveAspectFit`: longest side matches the tile, the other side letterboxes.
- Folder (and other) glyphs scale to the same square.
- Stroke is inverted against that scale so a large folder keeps the original grid-icon weight instead of fattening. `Glyph.strokeWidth` is settable for that one caller; list and rail stay on `Theme.strokeWidth`.

List and columns are unchanged.

## Files

- `ui/GridTile.qml`
- `ui/GridArea.qml`
- `ui/Glyph.qml`
- `ui/Theme.qml` (comment only: `grid.iconSize` is now the stroke-weight reference, not the painted slot)

Dogfooded on Omarchy against Pictures in grid view.

## Summary by CodeRabbit

* **UI Improvements**
  * Updated grid tiles to use square layouts with thumbnails that fill available tile width while preserving aspect ratio.
  * Adjusted glyph sizing and stroke weight to scale with tile dimensions.
  * Improved grid spacing and cell sizing for more consistent layouts.
  * Ensured hovered tiles and their labels appear above neighboring cells.
  * Added support for overriding the default glyph stroke width when needed.

### DIFF
--- a/ui/Glyph.qml
+++ b/ui/Glyph.qml
@@ -17,7 +17,9 @@ Item {
     readonly property real markSize: Math.min(root.maxSize, root.width, root.height)
     // Its own name: "scale" would shadow Item.scale, which qmllint flagged.
     readonly property real gridScale: root.markSize / root.grid
-    readonly property real strokeWidth: Theme.strokeWidth
+    // Grid units, then scaled with the path. A large slot overrides this so screen-space thickness
+    // can stay at the designed weight instead of growing with the mark.
+    property real strokeWidth: Theme.strokeWidth
 
     // The Shape is sized in grid units pre-transform; Scale's origin defaults to (0,0), so this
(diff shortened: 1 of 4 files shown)


Jijjy:
Addressed the CodeRabbit note: `GridArea.thumbPx` is now the tile's `slotSize`, so the painted slot and the reserved cell share one clamped value.

Also fixed names and hover tips painting through the next row (the cell only reserved one caption line while the label wraps to two; the tip's `z` was inside the delegate, so later tiles covered it).

PR #53, opened 2026-09-05
dd keeps the cursor in place, and shows what it's about to trash

W4HO-ham:
Two things that bugged me when trashing with dd.

1. After the trash, the whole list reloaded and the cursor jumped back to the top. Now the rows stay on screen and the cursor stays where it was, on whatever slid up into the gap, same as vim. If you trashed a selection it lands on the first row of the block. Paste and undo get the same treatment since they go through the same refresh.

2. The first d only showed a status line. Now it also tints the rows the second d will remove, using the theme's urgent color, so you can see what's about to go before you commit. The tint clears after the same 1.5s window the pair already used, or on any other key. Works in the list and grid views.

No backend changes. tests/js.sh is green and I added checks to nav.js and trash.js. I couldn't run tests/ui.sh here (no omarchy-drive), but I drove the window by hand and checked single row, last row, selection, emptying the directory, and undo.

## Summary by CodeRabbit

* **New Features**
  * Added clear visual indicators for rows and tiles targeted by the two-step trash action.
  * Trash-targeted items use error-colored highlights, borders, and markers.
  * Directory rows now show clearer names, links, dates, sizes, and icons.
  * Archive files open Flea’s preview when activated; directories continue navigating normally.

* **Bug Fixes**
  * Refreshing directories preserves cursor placement and applies pending selections safely.
  * Cursor placement remains consistent after deleting files or selections.

### DIFF
--- a/tests/js/nav.js
+++ b/tests/js/nav.js
@@ -24,11 +24,16 @@ function pane() {
         lockedMode: 0o40750,
         filterQuery: "scr",
         filterTyping: true,
+        pendingSelect: "",
+        pendingCursor: -1,
         cleared: 0,
         said: [],
-        sent: []
+        sent: [],
+        placed: []
     }
     p.clearSelection = function () { p.cleared += 1 }
+    p.setCursor = function (index) { p.cursorIndex = index; p.placed.push(index) }
+    p.join = function (base, name) { return base + "/" + name }
     p.message = function (text, isError) { p.said.push(text) }
     p.listArea = { primeSettle: function () {} }
     p.backend = {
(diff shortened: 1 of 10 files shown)


W4HO-ham:
Closing this one. The cursor fix landed on main in 1cdc890 and shipped in 0.3.0. I built it from the tag and tested it here, and dd leaves the cursor right where it belongs now, single row and block both. That was the part that bugged me, so I'm happy.

The tint on the first d was declined in the same commit, so there's nothing left here to merge.

Thanks for picking it up and doing it properly.

PR #60, opened 2026-09-06
Hand AppImages and system packages to AppShelf

maxmya:
`gio open` has no good answer for these. An AppImage has no desktop association on a fresh box, and a `.deb` resolves to an archive manager, which offers to unpack a package rather than install it. AppShelf is Omarchy's local application manager and shows the same review window for all four kinds, so Enter or a double-click goes there instead.

`src/appshelf.rs` decides what AppShelf handles — by content where the format has any, by name where it does not:

- **AppImages**: extension, or the Type 2 magic at offset 8.
- **Arch packages**: a `.pkg.tar*` name. There is nothing else to read, so `photos.tar.zst` deliberately does not match.
- **Debian packages**: extension, or an `ar` archive whose first member is `debian-binary`. The `!<arch>` magic alone also matches a static library, which must keep going to the desktop.
- **RPM packages**: extension, or magic.

None read past the first few bytes — AppShelf reads the file itself and refuses it with its own message when the guess was wrong.

When `appshelf` is not on `PATH` nothing changes: the file falls through to `gio open` and the desktop's associations still decide.

## Notes

- `src/appimage.rs` is renamed to `src/appshelf.rs`, since it now covers four formats rather than one.
- 5 unit tests; full suite passes (429).
- `tests/modes.sh` gains cases for `.deb`, `.rpm` and `.pkg.tar.zst` handoff plus a negative case for a plain `.tar.zst`. **These were not executed here** — the harness needs a root-owned `/home/flea-sandbox`.

## Summary by CodeRabbit

- **New Features**
  - Opening AppImages and Arch, Debian, and RPM packages now hands them to AppShelf when available.
  - Unsupported files continue opening with the system default handler.
  - AppShelf is now listed as an optional package for supported file types.

- **Bug Fixes**
  - Invalid command-line argument combinations now produce clear usage errors.

- **Documentation**
  - Added documentation describing supported file types and opening behavior.

- **Tests**
  - Added coverage for AppShelf handling and fallback behavior.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -3441,14 +3441,14 @@ Omarchy's local application manager and shows the same review window for all of 
 - **AppImages**: the `.AppImage` / `.appimage` extension, or an extensionless binary carrying the
   Type 2 AppImage magic `\x7fELF` and `AI\x02` at offset 8.
 - **Arch packages**: a name ending `.pkg.tar.zst`, `.pkg.tar.xz`, `.pkg.tar.gz`, `.pkg.tar.bz2`,
-  `.pkg.tar.lzo` or `.pkg.tar`. There is no magic to read — the file is an ordinary compressed
-  tarball — so the name is the whole claim, and `photos.tar.zst` deliberately does not match.
+  `.pkg.tar.lzo`, `.pkg.tar.lrz`, `.pkg.tar.lz4`, `.pkg.tar.lz`, `.pkg.tar.Z` or `.pkg.tar`.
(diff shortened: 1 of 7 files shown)


maxmya:
WIP

maxmya:
@coderabbitai resolve

maxmya:
@thisisgm  wdyt ?

maxmya:
btw this code was written and handled by me with agents ( claude , codex )

PR #78, opened 2026-09-07
fix: PDF preview shows an empty dark frame

DouglasdeMoura:
## The problem

Previewing a PDF — `l`/Space in the list, or the preview frame in the columns view — opens an empty dark frame. The document actually loads fine (the page count shows, the pager works), but the page itself doesn't appear, or only its text shows floating on the dark background.

ev-columns-pair

ev-overlay-pair

## Why

Qt's PDF page image sometimes arrives with a transparent background: the text is there, the white paper isn't. The dark frame behind shows through, and the whole thing reads as "empty". The app can't tell — the document reports Ready — only the pixels are wrong.

## The fix

One small change in `ui/PreviewPdf.qml`: paint the paper ourselves. A white rectangle sits exactly behind the page image, visible only while a document is actually shown. If Qt delivers the paper normally, the rectangle is simply hidden behind it; if not, it *is* the paper.

Both the Quick Look and the columns preview share this file, so one change covers both.

## Summary by CodeRabbit

* **UI Improvements**
  * Added a white page background behind visible PDF pages, providing a clearer paper-like appearance during document viewing.
  * The background is sized and centered to match the displayed page and remains hidden while the document is loading.

### DIFF
--- a/ui/PreviewPdf.qml
+++ b/ui/PreviewPdf.qml
@@ -45,6 +45,20 @@ Item {
         source: root.active && root.path.length > 0 ? Format.fileUri(root.path) : ""
     }
 
+    // The page's own paper, painted beneath the raster: on this box the rendered page can arrive
+    // with a transparent background, text drawn and paper not, and the dark frame behind showed
+    // through it (measured in the Quick Look and in the column, Vulkan and GL, with the document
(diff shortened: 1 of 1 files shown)


thisisgm:
Merged in 455b20d (v0.1.6) with your commit as its parent: the paper rectangle sits under PDF pages in every preview; thanks.

PR #81, opened 2026-09-07
Use Omarchy's regular text size for filenames and controls

### LINKED ISSUE #79: Proposal: use regular Omarchy body text for filenames and normal reading
# Proposal: regular Omarchy text size in Flea

Status: draft for review. No runtime changes implemented.

## Goal

Normal reading text in Flea should use Omarchy's regular body size, not a smaller
variant. Following the desktop should mean matching its regular reading size,
without requiring a Flea-specific zoom override.

Keep Omarchy's font family, theme colours and desktop scaling. Do not increase the
user's global text size or rewrite their saved Flea preferences to achieve this.

## Current behaviour

`ui/Theme.qml` exposes `bodySmall` and `caption`, but no regular `body` token.
While following Omarchy, both come directly from `qs.Commons.Style`. Omarchy's
`Color` singleton merges the theme's `shell.toml` with the watched user override at
`~/.config/omarchy/shell.toml` before applying it to `Style`.

With a base size of 12 and no per-token overrides:

| Surface | Current size | Proposed size |
|---|---:|---:|
| List filenames and sidebar labels | 11 px (`bodySmall`) | 12 px (`body`) |
| Grid filenames | 10 px (`caption`) | 12 px (`body`) |
| File size, date, permissions and kind | 10 px (`caption`) | 12 px (`body`) |
| Normal controls, inputs and menu labels | Usually 11 px | 12 px (`body`) |

These are logical pixels, before Qt/compositor scaling. At base 14, regular body
text normally becomes 14 px, not Flea's current 13 px names and 12 px metadata.
A theme's explicit `body` override must still win while following Omarchy.

## Typography contract

- **Regular body:** filenames in every view and the picker, sidebar entries,
  paths, tabs, column labels and file metadata; editable values and their labels;
  menu actions, settings controls, status/error messages and preview reading text.
- **Caption:** explicitly secondary hints, shortcut annotations and compact
  section headings. A filename or file detail is never a caption merely to fit.
- Preserve existing heading emphasis where appropriate. Distinguish secondary
  metadata primarily through colour, not reduced text size.
- Keep genuine `bodySmall` and `caption` tokens for intentional secondary uses.
  Do not silently redefine either token to mean `body`.

### 1. Introduce the missing regular token

Add `Theme.font.body` in `ui/Theme.qml`:

```qml
readonly property int body: root.overridden ? root.baseSize : Style.font.body
```

Keep `family: Style.font.family`. Following uses the actual resolved Omarchy body
size, including theme/user token overrides. An explicit Flea size uses the pinned
base at a 1:1 ratio. Existing size stops, persistence and reset behaviour stay intact.
Expose `body` through `Theme.tokens()` for runtime verification.

### 2. Migrate by text role, not by global replacement

Start with the shared browser components: `MatchText.qml`, `Row.qml`,
`GridTile.qml`, `SidebarRow.qml` and `RenameField.qml`. Display and rename must
use the same size. Include the columns view and chooser consumers of these
components so switching views cannot shrink the same filename.

Then audit chrome, menus, dialogs, settings and preview components against the
contract above. For imported Omarchy controls, use their supported font properties
where needed; do not edit `/usr/share/omarchy/` or its symlinked components.

Each remaining small-text use must have a secondary role. Do not replace references
that size an icon, indicator or hit target as though they were text bindings.

### 3. Make geometry accommodate the text

Column measurement currently uses `bodySmall`, as do several width and height
formulas. Measure normal text columns using `body` and update their shared widths
for both header and rows. Review picker dates, kind/name minimum widths, grid labels,
rename editors and menu widths as part of the same change.

Retain existing padding and icon sizes where they still fit. Increase a container
where real text metrics require it rather than shrinking the font back down.
Do not globally rescale all icons and spacing merely because body text grew.
Test explicit Omarchy `body` overrides as well as base-size changes: the base and
body token need not be equal.

The viewport/window calculation must use the final rendered row height. Preserve
the integer view model and viewport-scoped metadata/thumbnail requests.

## Acceptance checks

- At base 12 with no overrides, normal reading text renders at 12 logical px;
  at base 14 it renders at 14. Assert actual text-item font properties, not only
  the presence of the new token.
- Following honours a theme/user `body` override and live Omarchy size changes.
  A pinned Flea size stays pinned; Ctrl+Shift+0 restores the desktop value.
- Exercise all seven existing Flea size stops and the real monitor scale. No double
  multiplication of compositor scale is introduced.
- Compare screenshots of list, grid, columns, picker and settings with Omarchy
  controls using `Style.font.body`. Check clipping, alignment, long names, rename
  focus, menu fit and narrow windows. Existing responsive column hiding remains.
- Update `tests/js/textsize.js`, the relevant `tests/ui.sh` checks and
  `tools/flea-metrics-gate` to express the approved typography/geometry contract.
  Preserve tests for the smaller tokens that remain; do not just raise every
  expected number. Verify that reverting a primary label to a smaller token fails.
- Run `tests/run-all.sh`, then the relevant live UI checks it does not run, plus
  the QML lint and file-budget gates. Recheck thumbnail viewport bounds if row
  geometry changes. Record any changed layout baselines explicitly.

## Trade-off and decision

Larger file details consume more horizontal space, so narrow windows may show fewer
optional columns or elide sooner. Rows may also grow if needed to avoid clipping.
That is preferable to making ordinary information smaller than the desktop's regular
text. Actual layout and performance impact remains to be measured, not assumed.

Proposed decision: adopt regular body text for ordinary reading across Flea, keeping
smaller text only for the explicit secondary roles above. Implement after review.

aholbreich:
Fixes #79.

Flea looked a bit too small next to the rest of Omarchy. With the desktop set to 12px, list filenames were 11px and grid filenames and file details were 10px. Settings still said “Effective 12px”, which was confusing.

This uses Omarchy's regular body size for filenames, file details, controls and messages. It isn't a global zoom: hints, shortcut annotations and section headings stay smaller. Icons and most spacing stay as they were. The aim is to make the text you actually read more comfortable without enlarging the whole interface.

Column widths now follow the larger text. Rows only grow when the font needs more room, and grid tiles reserve space for both lines of a filename. Settings shows “Base” and “Reading text” separately, since a theme can set its body size independently of its base size. Existing font choices, size overrides and monitor scaling still work the same way.

Built and installed the package locally. It looks better without everything getting bigger.

### Testing

- 437 Rust tests passed in both debug and release.
- JavaScript, keymap, QML lint and file-budget checks passed.
- New offscreen tests check actual component fonts at 1× and 2× scaling, all seven size stops, and theme/user overrides. Reverting filenames to `bodySmall` makes them fail.
- Reviewed offscreen captures of list, grid, columns and Settings.

The full runner was **19/24**, not a clean pass. Media fixtures were missing (`protocol`, `thumbs`, `media`), trash tests ran on `/tmp` (`ops`), and the state-file kill sweep missed its timing threshold (`uistate`). Compositor-driven UI tests weren't run because `omarchy-drive` isn't available here.

## Summary by CodeRabbit

* **New Features**
  * Updated typography across file listings, previews, dialogs, settings, and status areas for more consistent body-text sizing.
  * Added expanded Settings sections for View, Places, Preview, and About, including keyboard previews and improved navigation.
  * Added keyboard navigation, focus handling, scrolling, paging, and clearer controls to the PDF viewer.
  * Improved picker workflows with filtered listings, save-path checks, asynchronous validation, and clearer error handling.
  * Added expandable status details with retry, undo, dismiss, and transfer controls.
* **Bug Fixes**
  * Improved focus restoration, layout sizing, text wrapping, and display scaling behavior.


thisisgm:
Partly in v0.1.6 (784da46): running text now draws at Omarchy's regular body size, following Omarchy or the pinned stop; grid and metadata captions keep their size by design, so this stays open for the rest.

PR #84, opened 2026-09-08
feat(devices): list internal HDD volumes in the rail

tharinfernando:
An internal HDD automounted under /run/media never appeared in the Devices rail: collectVolumes only emitted rm partitions, so a second fixed drive (rm=false, tran=sata) was neither the machine row nor a volume. Nautilus lists it, Flea did not. Measured on a WDC WD10SPZX (931.5G, sda1 at /run/media/tharin/<uuid>).

## What changes

- ui/js/Mounts.js: parseDevices takes home and builds the udisks2 prefix /run/media/<user>/; an internal partition is a row when it is automounted there or unmounted with nothing stacked on it (LUKS/LVM containers, /boot and swap stay excluded). The parent disk's rm covers children, since newer lsblk reports per-partition rm=false even on sticks. Volume rows carry removable, railMenu offers Eject only to removable ones, and sameEntry compares the flag.
- ui/DeviceMounts.qml: passes HOME through and carries removable into entries (plus the missing import Quickshell the env call needs).
- ui/SidebarRow.qml: an unmounted device volume draws a muted chevron-right in the indicator slot instead of the dim square; the slot is unchanged, so no label shifts and no geometry seam moves. Mounted rows keep the dot.
- tools/flea-file-budget: known-cap exception for ui/js/Mounts.js and tests/js/mounts.js, both already at the cap.

## What it deliberately does not do (v1)

- No Eject/Unmount on internal volumes: eject sends gio mount -e, which is removable-media semantics. The row opens and nothing else.
- Internal partitions mounted outside /run/media (e.g. /mnt/data) still do not list.

## Verification

- ./tests/js.sh mounts: 148 checks, 0 failed (21 new, built off the live sda shape).
- ./tests/js.sh (all suites), ./tests/budget.sh, ./tools/flea-qmllint-gate: green, 0 regressions.
- Live on the box: the automounted sda1 UUID path renders as a mounted row; unmounting it leaves the row with the arrow, and clicking mounts and opens through the existing gio mount -d flow. USB, zram, /boot and the LUKS stack verified unchanged.
- tests/ui.sh eject not run: it drives a stubbed lsblk and needs the display session.

## Summary by CodeRabbit

- **New Features**
  - Internal hard-drive volumes mounted under the user’s media directory now appear in the device sidebar.
  - Device entries distinguish removable storage from internal volumes.
  - Unmounted volumes display a muted mount indicator to clarify their state.

- **Bug Fixes**
  - Eject actions are now available only for removable volumes.
  - Unsupported system mounts and encrypted container layers are excluded from the device list.


thisisgm:
Landed in v0.3.0 reworked onto the rail board with credit; closing in favour of that commit, thank you.

PR #86, opened 2026-09-08
fix: undo brings a move back across filesystems

nixfred:
## The bug

Move a file from a real disk onto a USB stick, another partition or tmpfs, then press undo. The reversal answered `{"t":"error","where":"rename",...,"msg":"Invalid cross-device link (os error 18)"}` and the file stayed where it was moved to. Because `Journal::undo` pops the entry before reversing it, the entry was also spent: the next undo reached an older operation.

The forward path already knows a cross-filesystem move is a copy and a remove (`copyfile.rs`'s `move_any` on `EXDEV`). The reverse path goes through `renamecompat::rename_path`, whose copy fallback only fired for an rclone `EINVAL` or a WebDAV `EIO`.

## The fix

`needs_copy_fallback` answers true for `EXDEV`, so the reversal takes the same `copy_then_remove` arm the two measured mounts use: exclusive create, partial cleanup on a failed copy, and the `rename-kept` rule when the source will not go. `docs/protocol.md`'s `undo` section says a move that crossed filesystems copies back the way it went out.

## Tests

- Unit test pins the errno in `renamecompat.rs`, alongside the existing WebDAV and rclone predicate tests.
- `tests/ops.sh` gains a scenario that moves a file and a nested tree from the fixture root onto a `/dev/shm` tmpfs root and undoes them, using the same self-marked tmpfs root pattern `tests/drag.sh` R7 uses, with the device check that proves the two really are different filesystems. On the unfixed binary the scenario fails four checks: the undo answers a rename error and both sources stay gone.

## Summary by CodeRabbit

- **Bug Fixes**
  - Cross-filesystem moves now complete successfully by copying items to the destination and removing the originals.
  - Undoing a cross-filesystem move now restores the item to its original location and removes the destination copy.

- **Tests**
  - Added coverage for cross-filesystem file and directory moves, including undo behavior and cleanup.

### DIFF
--- a/docs/protocol.md
+++ b/docs/protocol.md
@@ -487,7 +487,8 @@ through the same call a `rename` does, so its failure answers `rename` or `renam
 
 **The journal is an in-memory ring of the last 50 completed operations and is not persisted**, so it
 does not survive a restart. Each kind reverses as follows: a rename or a move renames back (still
-refusing to clobber, because something may occupy the old name by now), a copy or a duplicate removes
+refusing to clobber, because something may occupy the old name by now; a move that crossed filesystems
+copies back and then removes, the way it went out), a copy or a duplicate removes
 what that operation created, and a trash restores through `gio trash --restore` using the URI captured
 when it was trashed. A `mkdir` removes the folder it made only while it is still empty: a folder the
 user has filled since is theirs, so that reversal answers an `error` line, leaves it and its contents in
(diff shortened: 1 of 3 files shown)


thisisgm:
Superseded by `5c3b594` (`v0.2.0`): undo already uses the exclusive-copy fallback for cross-device renames; this PR was not merged.

PR #87, opened 2026-09-08
fix: a surrogate pair on the wire names its character

nixfred:
## The bug

`json::field_str` had its own copy of the JSON escapes, and its `\u` arm called `char::from_u32` on each four-digit escape alone. A surrogate half is not a scalar value, so `\ud83d` made the whole field answer `None`, and the callers read that as "absent":

- `{"c":"list","path":"/x/📁 Work","first":1}` listed `""` and answered a `scan` error with an empty path.
- `{"c":"trash","paths":["/a","/x/📁.txt"]}` dropped the second element inside `field_str_array` and answered `{"t":"trashed","ok":1,"failed":0}` with no error line for the file it never touched.

The shipped QML client emits raw UTF-8, so it never hits this, but any ASCII-safe serializer does (Python's `json.dumps` default, `jq -a`), and `docs/protocol.md` calls every string on the wire JSON. `src/jsonstring.rs`, the reader `jsondoc.rs` uses for the state file, already pairs surrogates correctly, so the tree had two readers that disagreed on valid JSON.

## The fix

`field_str` finds the opening quote as before and hands the string to `jsonstring::parse_string`. One reader for both sides: a pair is the one character it names, and a half with no partner is the replacement character, the same answer the document reader gives. `field_str_array` goes through `field_str`, so array elements are covered by the same change.

## Tests

- New unit test in `json.rs` pins the pair in a `path` field, the pair inside a `paths` array element keeping its place, and the lone-half replacement. On the old reader it fails with `left: None`.
- `cargo test`: 442 passed, zero warnings in debug and release. `tests/ops.sh` passes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved JSON string decoding for standalone fields and array elements.
  - Correctly preserves non-BMP characters represented by surrogate pairs.
  - Applies consistent replacement-character handling for unmatched surrogates.
  - Ensures string parsing behaves consistently across whole documents and individual values.

### DIFF
--- a/src/json.rs
+++ b/src/json.rs
@@ -44,38 +44,18 @@ fn value_start(line: &str, key: &str) -> Option<usize> {
     None
 }
 
-// Sample input: "/tmp/two\nlines" or "café", the value only, quotes included.
+// Sample input: "/tmp/two\nlines" or "café", the value only, quotes included. The string itself is
+// read by src/jsonstring.rs, the one reader the whole-document side already has, so a \ud83d\udcc1
+// pair lands as the one character it names on this wire too: this scanner had its own copy of the
+// escapes that answered None to any surrogate, and a path holding one then read as absent, which
+// listed "" and trashed nothing without an error line to say so.
 pub fn field_str(line: &str, key: &str) -> Option<String> {
     let start = value_start(line, key)?;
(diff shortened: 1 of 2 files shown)


thisisgm:

PR #88, opened 2026-09-08
fix: a path already gone is not trashed, and does not block undo

nixfred:
## The bug

`trash::trash` decided which paths went by stat-ing each one after the `gio trash` call: still on disk means failed, anything else means trashed. A path that was already gone before the call, which a stale listing can name until the `changed` re-read lands, therefore counted as `ok` and was journaled as a `Trashed` step with an empty URI.

`Journal::undo` reverses newest first and stops on the first failure, so with `paths:["real.txt","already-gone.txt"]` the reply was `trashed ok:2 failed:0`, and the undo answered "this item was trashed without a trash entry, so it cannot be restored". `real.txt` stayed in the trash and the entry was spent, so the next undo reached an older operation. With the order reversed the real file was restored and then the error fired, so the outcome depended on row order.

## The fix

The paths are partitioned before `gio` runs. One that is not on disk is counted in `failed` and never becomes a step; a batch with nothing present never runs `gio` at all. The rule for a path still on disk after the call is unchanged, and so is the corner where a file went but `gio` listed no entry for it. `docs/protocol.md`'s `trashed` section says so.

## Tests

- Unit test in `trash.rs`: two missing paths answer no entries and `failed:2`, without running `gio`.
- `tests/ops.sh` gains a scenario trashing a real file beside a path that never existed: `ok:1 failed:1`, and the undo restores the real file with its bytes. On the unfixed binary it fails three checks: the reply reads `ok:2`, the undo answers an error, and the file stays in the trash.

### DIFF
--- a/docs/protocol.md
+++ b/docs/protocol.md
@@ -837,7 +837,8 @@ Example: `{"t":"trashed","ok":1,"failed":0}`
 
 Counts only. Unlike `transferitem` there is no per-path error text, because trash is one `gio` call for
 the batch and its exit status cannot attribute a failure to a single path; a path that is still on disk
-afterwards is counted in `failed`.
+afterwards is counted in `failed`, and so is one that was not on disk before, which a stale listing can
+name: nothing this call did not do is journaled, so `undo` reverses exactly the `ok` ones.
 
 ### renamed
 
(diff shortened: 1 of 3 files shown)


thisisgm:

PR #89, opened 2026-09-08
fix: a search row's kind is read from its name, not its path

nixfred:
## The bug

`mime::Db::lookup` is documented as taking a file name, never a path. `rows_line` and `thumb_rows` pass it `listing.name(i)` verbatim, which for a `search` or `listpaths` listing is the path relative to the base (`docs/protocol.md`: "each match is pushed as its path relative to `path`"). The by-name globs (`makefile`, `cmakelists.txt`, `meson.build`, `dockerfile`, ...) are keyed on bare names, so they never matched a nested row.

Driven against the backend on a tree holding `Makefile`, `src/Makefile` and `src/CMakeLists.txt`, a search answered:

```
"kinds":["Makefile build file","Data","Plain text document"]
```

with `src/Makefile` as "Data" and `src/CMakeLists.txt` as "Plain text document" via the `.txt` suffix. The same files listed directly answer "Makefile build file" and "CMake source code". The picker's Recent (`listpaths`) has the same defect, and two smaller ones ride along: a directory component's dot could offer a suffix of its own (`v1.2/notes`), and a hidden file below the base (`src/.jpg`) no longer read as hidden because its dot was not at offset 0.

## The fix

`lookup` takes the last `/`-separated component itself, so every caller is covered with one change: the rows line, the thumbnail request's declared-MIME check, and `peek`'s icon.

## Tests

- Unit test in `mime.rs` for the nested by-name hit, a nested suffix, a directory's dot, a hidden file below the base, and a name with no extension under a directory that has one.
- `tests/protocol.sh` holds a search row to the kind `list` gives the same nested Makefile. On the old binary it fails with `expected "kinds":["Makefile build file"]`, `actual "kinds":["Data"]`.
- `cargo test`: 442 passed, zero warnings in debug and release. `tests/ops.sh` passes. The three `protocol.sh` thumbnail checks that need the 4.7 GB media fixture were not runnable on this box, before or after.

### DIFF
--- a/src/backend/mime.rs
+++ b/src/backend/mime.rs
@@ -70,8 +70,12 @@ impl Db {
         Db { by_suffix_cs, by_suffix, by_name_cs, by_name }
     }
 
-    // Takes a file name, never a path: a directory component must not be read as an extension.
+    // Answers for the last component alone. A search or listpaths row is a path relative to the listing's
+    // base, and looked up whole the by-name tables never matched it: src/Makefile listed as Data and
+    // src/CMakeLists.txt as Plain text document, while a directory component's dot could offer a suffix
+    // of its own and a hidden file below the base could stop reading as hidden.
     pub fn lookup(&self, name: &str) -> Option<&str> {
+        let name = name.rsplit('/').next().unwrap_or(name);
         let lower = name.to_lowercase();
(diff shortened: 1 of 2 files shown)


thisisgm:

PR #90, opened 2026-09-08
fix: a JPEG is measured behind a long EXIF segment

nixfred:
## The bug

`imagesize::dimensions` reads an 8 KiB probe and `jpeg()` walked the segment chain inside that buffer only. A camera or phone JPEG carries an APP1 (EXIF) segment of 20 to 60 KiB holding the embedded preview before the SOF marker, so `i += 2 + len` left the buffer and the walk ended with `None`. The comment claimed the probe was "enough ... on every file this box produces".

Driven against the backend with a synthetic JPEG holding a 20,000-byte APP1 then SOF0 640x480, and the same file with a 4,000-byte APP1:

```
before: row 0 (long APP1)  "w":0,"h":0   "lines":1
        row 1 (short APP1) "w":640,"h":480
after:  row 0              "w":640,"h":480 "lines":0
        row 1              "w":640,"h":480
```

The `lines:1` is the knock-on: `metareq` skips the line count only when the header parsed as an image, so with the `text` hint the photo was counted for newlines.

## The fix

The JPEG walk runs over `Read + Seek`: it reads each marker and length and seeks past the payload, a few bytes per segment however long the EXIF runs, so nothing is decoded and nothing large is read. It also handles what the byte walker mis-stepped on: any number of `FF` fill bytes before a marker, the length-less markers (TEM, RSTn, a repeated SOI), and a scan or EOI before any frame header, which ends the walk instead of scanning into entropy-coded data. `from_header` keeps its byte interface by running the same function over a `Cursor`, so the two cannot disagree. The other formats are unchanged and still answer from the probe.

## Tests

- A JPEG on disk with a 20,000-byte APP1 measures 640x480, and the same bytes cut at the probe length answer `None` from `from_header`, which documents why the file is walked.
- Fill bytes, a lone restart marker, a scan before the frame, and a non-marker byte where one is due.
- `cargo test`: 443 passed, zero warnings in debug and release.

### DIFF
--- a/src/backend/imagesize.rs
+++ b/src/backend/imagesize.rs
@@ -1,14 +1,18 @@
 // Pixel dimensions read from a file header, because the preview column names them and the wire does
-// not carry them. Only the first bytes are read: no image is ever decoded to answer this.
+// not carry them. Only the header is read: no image is ever decoded to answer this, and a JPEG is
+// walked segment by segment with a seek over each payload, never read through.
 use crate::backend::regfile::open_regular;
-use std::io::Read;
+use std::io::{Cursor, Read, Seek, SeekFrom};
 use std::path::Path;
 
-// Enough for a JPEG's segment chain to reach its first frame header on every file this box produces.
+// The fixed-offset formats answer inside this; a JPEG is walked from the file instead, because its frame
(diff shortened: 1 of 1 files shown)


thisisgm:

PR #91, opened 2026-09-08
fix: a tab switch spends its pending order on the first rows reply

nixfred:
## The bug

`Tabs.apply` on a switch that re-lists records the tab's order in `pendingSortBy`, and `Backend.list` resets the order to name ascending, which is the order most tabs record. `Tabs.applyPending`, run by `PaneWire.onRows` on every rows reply, cleared `pendingSortBy` only inside the branch that found the order different and asked for it. A pending `"name"` that already matched was therefore never cleared and stayed pending for the life of that tab.

The next time the user sorted in that tab (`s`, `S`, or a header click), the rows reply answering their sort ran `applyPending` again, saw `"name" !== "size"`, and sent `sort("name", false)`: the header flickered and landed back on Name. The second press stuck, because the revert branch is the one that clears the field.

## The fix

The pending order is spent on the first rows reply whichever way the comparison goes. If it differs it is asked for as before; if the fresh listing already has it, nothing is sent and the pending cursor is restored on that same reply instead of the next one.

## Tests

- `tests/js/tabs.js` drives a two-tab switch: the pending order is recorded, the first rows reply spends it without a sort request and restores the cursor, and a later user sort survives the rows reply that answers it. On the old code two checks fail: `got "name|"` and `got "name:false"`.
- `./tests/js.sh`: 1,832 checks, 0 failed. qmllint gate passes.

### DIFF
--- a/tests/js/tabs.js
+++ b/tests/js/tabs.js
@@ -211,4 +211,21 @@ function run(check) {
           pending.sorted.join(",") + "|" + pending.cursorIndex, "size:true|4")
     Tabs.applyPending(pending)
     check("the next rows reply restores the cursor", pending.cursorIndex, 9)
+
+    // A switch that re-lists records the tab's order as pending, and a fresh listing is already name
+    // ascending, which is the order most tabs record: left pending because it matched, it reverted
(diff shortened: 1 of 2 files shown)


thisisgm:

PR #92, opened 2026-09-08
fix: a cursor row the filter hides is not a row to act on

nixfred:
## The bug

`Filter.apply` moves the cursor onto the first row still standing only when the query left one standing. A query that matches nothing (the list shows "Nothing matches z") leaves the cursor on the row it was on, which is now hidden. `Ops.targetIndices` falls back to `[pane.cursorIndex]` once `prune` has emptied the selection, and `Nav.openCursor` and `Ops.startRename` read the same row, so over an empty filter:

- `dd` trashed the hidden row, and `y` / `x` / Move to Dropbox took it.
- Enter opened it, navigating away from the filtered listing.
- `r` set `renamingIndex` with no delegate to draw the editor. That holds `PaneWire.watchBusy` until navigation, and the editor pops open over the row when Escape clears the filter.

`prune()`'s own comment states the rule: a row you cannot see is one you can act on by accident, and trash, cut and copy take their target with no confirmation. The selection was held to that rule; the cursor fallback was not.

## The fix

`Filter.cursorShown(pane)` answers whether the cursor is on a row the filter draws (always true with no filter). `targetIndices` answers no rows for a hidden cursor, `startRename` opens no editor, and `openCursor` says "That row is hidden by the filter." rather than opening it. Nothing changes for a visible cursor or for a standing selection.

## Tests

- `tests/js/filter.js` drives all three over a `zzz` filter on the fixture: no target, no editor, no open; and the visible-cursor and no-filter cases keep their targets. On the old code the checks fail three ways, the last by opening the hidden directory.
- `./tests/js.sh`: 1,833 checks, 0 failed. qmllint gate passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented actions from targeting rows hidden by an active filter.
  * Renaming no longer starts when the cursor is on a hidden row.
  * Opening a hidden cursor row now displays a message instead of navigating or opening it.

### DIFF
--- a/tests/js/filter.js
+++ b/tests/js/filter.js
@@ -1,4 +1,6 @@
 .import "../../ui/js/Filter.js" as Filter
+.import "../../ui/js/Nav.js" as Nav
+.import "../../ui/js/Ops.js" as Ops
 .import "filterfixture.js" as Fixture
 
 // The fixtures and the stub pane live in filterfixture.js, so this file stays checks.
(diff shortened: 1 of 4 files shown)


thisisgm:

PR #93, opened 2026-09-08
fix: a tab action out of a search re-lists the search's own scope

nixfred:
## The bug

`Tabs.dropOverlay` clears `searchMode` but leaves `pane.path` on the scope the walk was given and `rows` / `total` on the walk's results. `openNew`, `selectAt` and `closeAt` then re-list only when the destination path differs from `pane.path`. `Search.close`, the Escape path, always re-lists.

A search begun in home walks home (`Search.scopeRoot`), so the scope is exactly where the user was:

1. Flea open at `~`, press `f`, search, Enter, then `t`: the new tab labelled Home shows rows like `Downloads/report.pdf` under the normal column header, with no search chrome.
2. Tab 1 at `~`, tab 2 at `~/Downloads`; run a home-wide search in tab 2 and press `1`: the same.
3. Close a searching tab onto one standing on the scope: the same.

The backend drops the inotify watch for a `search` (its listing is not a directory), so no `changed` line ever repairs it; only navigating away does.

## The fix

`dropOverlay` answers whether it dropped a results listing. `openNew` re-lists on that as well as on a path difference, and `selectAt` and `closeAt` pass it to `apply`, whose "same path, nothing to re-list" shortcut no longer applies when the rows on the pane are a walk's.

## Tests

- `tests/js/tabs.js` drives all three paths with the scope equal to the destination and requires the re-list. On the old code all three fail with `got ""`.
- `./tests/js.sh`: 1,831 checks, 0 failed. qmllint gate passes.

### DIFF
--- a/tests/js/tabs.js
+++ b/tests/js/tabs.js
@@ -60,6 +60,35 @@ function run(check) {
     check("carrying no cursor from the search's own listing", searching.tabs.items[1].cursorIndex, 0)
     check("and no selection from it either", searching.tabs.items[1].selected.length, 0)
 
+    // A search begun in home walks home, so the scope IS where the user was: dropOverlay cleared the
+    // search and the path test found nothing to re-list, and the walk's rows stayed up under the new
+    // tab's ordinary header, with the watch dropped for the search so nothing ever refreshed them.
+    var scoped = pane("/home/gm")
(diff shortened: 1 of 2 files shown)


thisisgm:

PR #94, opened 2026-09-08
fix: a same-directory tab switch restores only a selection nothing renumbered

nixfred:
## The bug

`Tabs.apply` treats two tabs on the same directory as the one switch that re-lists nothing, and restores the snapshot's row-index selection verbatim. `ui/js/Selection.js` is a set of row indices, and its own rule is that a new listing clears them, because an index into a changed directory names another file. Two things renumber the rows under a hidden tab and neither was checked:

- **A different sort order in the other tab.** The same-path branch re-sorted (`backend.sort` + `window(0)`) but skipped the reset `Sort.resort` performs for exactly that reorder: the selection and the thumbnail and directory-size caches, all keyed by row index, stayed over a reordered listing.
- **A re-read while the tab was hidden.** Tabs A and B on `~/Downloads`; in A select rows 3 and 4; press `2`; in B press `dd` on row 0, or let another program create a file and the watch re-read; press `1`. Rows 3 and 4 are restored, which are now the files that were 4 and 5. The status bar reads "2 selected" over the wrong files, and the next `dd` would take them.

## The fix

The snapshot records `Backend.listRequests`, the counter every `list` bumps, including the watch's re-read and the refresh after a write. The restore happens only when it has not moved; otherwise the selection is cleared, which is what every other switch already does. The re-sort branch runs the same reset `Sort.resort` does: caches emptied, selection cleared.

## Tests

- `tests/js/tabs.js`: unchanged directory restores the selection; one re-read while hidden carries none; a re-sort drops it and the two caches. On the old code three checks fail, restoring `2,3` in both stale cases.
- `./tests/js.sh`: 1,832 checks, 0 failed. qmllint gate passes.

### DIFF
--- a/tests/js/tabs.js
+++ b/tests/js/tabs.js
@@ -90,6 +90,39 @@ function run(check) {
           moved.tabs.pendingSelected === undefined, true)
     check("and it did re-list, which is what clears the selection", moved.listed.join(","), "/tmp/b")
 
+    // The same directory in two tabs is the one switch that re-lists nothing, and the selection was
+    // restored on the strength of that alone. Two things renumber the rows underneath a hidden tab:
+    // the other tab holding a different order, and a re-read, a dd in the other tab or the watch's
+    // own, while it was hidden. Backend.listRequests counts the second; the sort is the first.
+    function twin(order) {
+        var p = pane("/tmp/same")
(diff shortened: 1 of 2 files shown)


thisisgm:

PR #95, opened 2026-09-08
fix: the tilde stands for home, not for a name starting with it

nixfred:
## The bug

`Format.tilde(path, home)` tested `path.indexOf(home) === 0`, a bare string prefix. `/home/gmx` is a sibling of `/home/gm`, not inside it, but it came back as `~x` and `/home/gmx/Work` as `~x/Work`. `Tabs.label` and the search strip's scope both go through `tilde`, so the tab strip and the header named a sibling directory as home's. `Nav.crumbs` had already worked around this for the path bar with its own whole-component test and a comment saying why.

## The fix

The test is home itself or home followed by a separator, the same test `crumbs` and `Search.scopeRoot` make. The `crumbs` comment now says the three agree rather than that `tilde` is wrong.

## Tests

- `tests/js/format.js`: the sibling and its child keep their own form. On the old code both fail with `~x`.
- `./tests/js.sh`: 1,830 checks, 0 failed. qmllint gate passes.

### DIFF
--- a/tests/js/format.js
+++ b/tests/js/format.js
@@ -10,6 +10,11 @@ function run(check) {
           Format.tilde("/usr/share/omarchy", "/home/gm"), "/usr/share/omarchy")
     check("an unknown home leaves every path alone",
           Format.tilde("/home/gm/x", ""), "/home/gm/x")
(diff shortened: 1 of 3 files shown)


thisisgm:

PR #96, opened 2026-09-08
fix: a file of exactly the line budget is counted whole

nixfred:
## The bug

`count_lines` returned `partial: true` as soon as the bytes read reached `LINE_BUDGET`, without knowing whether anything lay past it. A file of exactly 1,048,576 bytes had been read whole, but the early return skipped the `+1` for an unterminated last line and stated the count as a floor:

| file | before | after |
|---|---|---|
| 1,048,576 bytes of `a`, no newline | `lines:0, partial:true` | `lines:1, partial:false` |
| 1,048,575 bytes of `a`, no newline | `lines:1, partial:false` | unchanged |

## The fix

Reaching the budget asks for one more byte. Zero bytes back means the file was read whole and the normal ending applies; a byte back means there is more, and the count is the floor it always was.

## Tests

- Unit test: a file of exactly the budget ending in a newline counts one whole line; the same file plus a chunk and a byte is one line and partial. On the old code the first assertion fails with `left: (1, true, false)`.

### DIFF
--- a/src/backend/linecount.rs
+++ b/src/backend/linecount.rs
@@ -46,7 +46,14 @@ pub fn count_lines(path: &Path) -> LineCount {
         }
         read += n as u64;
         if read >= LINE_BUDGET {
-            return LineCount { lines: newlines, partial: true, failed: false };
+            // Reaching the budget is not passing it: a file of exactly the budget has been read whole,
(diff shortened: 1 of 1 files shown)


thisisgm:

PR #97, opened 2026-09-08
fix: the view the window is left on is the view the next launch opens on

DouglasdeMoura:
## Problem

Switching the view (columns, grid) is session-only: a restart opens on the list view whatever was on screen. Every other file manager on Linux persists the view mode — Nautilus, Dolphin, Thunar, pcmanfm and nemo all do.

This tree has carried the schema for it since the 0.1.4 handoff: `src/uischema.rs` validates `view` as `list`/`columns`/`grid`, `flea --ui-state '{"view":"columns"}'` accepts and merges it, and `src/uistore.rs`'s tests drive exactly that patch. The window just never read the key at startup and never wrote it on a change — `ui/Pane.qml`'s `viewMode` was hardcoded to `"list"` and the chrome's buttons and the ctrl-1/2/3 chords assigned it in memory. This PR wires the two ends together. No schema change, no wire change.

## The change

- `ui/ViewState.qml` — `view` reads the stored word and falls back to `"list"` for anything this build cannot draw, the way an unparsable `columns` set does; `setView` is the writer, through the same `changeKey` → patch machinery the settings already ride.
- `ui/Pane.qml` — `viewMode` starts on the stored view, and `setView(mode)` is the one user view change: it rebinds `viewMode` and persists.
- `ui/shell.qml` — the chrome's three buttons route through `pane.setView`.
- `ui/js/Focus.js` — the ctrl-1/2/3 chords route through the same seam, so a chord and a button cannot persist two different answers. `owe()` sends nothing for a value the window already holds, so a repeated chord is not a second write.

Two deliberate edges:

- **A tab switch does not rewrite the stored view.** `ui/js/Tabs.js`'s per-tab snapshot restores a tab's own view straight onto the property without a patch, because switching tabs is not choosing a view. Only an explicit user view change persists.
- **An unknown stored word draws the list view** (and the launch settle normally renames it back to the default, per the state file's per-key fallback). A file this window could not read is drawn as it was read, same as `columns`.

## Tests

- `tests/js/focus.js` — the view chords now assert `setView` was called. The old assertion read only the `viewMode` property, so a chord that switched the screen without persisting would have passed it; the mock records what reaches the seam.
- `tests/ui.sh case_viewpersist` — the real window: the ctrl-2 chord writes the file, a restart opens on the stored columns, the chrome's grid button does the same from the view the restart came up on, and a hand-edited `"view":"banana"` opens as the list view and is renamed back by the settle. `XDG_STATE_HOME` points inside the fixture root for the whole case. Run as `./tests/ui.sh viewpoints`.
- Headless, all green on this tree: `tests/js.sh` (1828 checks, 0 failed), `tests/keymap-gen.sh`, `tests/budget.sh`, the qmllint gate (0 regressions; the two new findings beside the documented false positives are the two QML files this change grew), `cargo test` (441 passed), both cargo profiles warning-free. `ui/Pane.qml` goes to 407 and stays a listed cap exception (it was 403); `ui/ViewState.qml` goes to 265, a soft-budget warning.
- Verified live on the window: stored columns boots in columns; Ctrl+3 flips `viewMode` and rewrites `ui.json`; a restart opens grid; `{"view":"banana"}` opens list and the settle rewrites it.

## One testing note

`./target/release/flea --gui` alone resolves the UI as `FLEA_UI` → packaged `/usr/share/flea/ui` → dev tree (`src/paths.rs`), so with the pacman package installed a dev launch runs the **packaged** QML unless `FLEA_UI` is set. The README dev loop already launches with `FLEA_UI="$PWD/ui"`; worth keeping in mind when testing this branch by hand.

## Summary by CodeRabbit

- **New Features**
  - Listing view preferences now persist across application restarts.
  - Supported views include list, columns, and grid; invalid saved values fall back to list view.
  - Toolbar controls and keyboard shortcuts consistently save view selections.
  - Grid layouts support improved keyboard navigation, selection, filtering, and actions.
  - Added share-browser keyboard actions and dual-pane focus switching.
  - Restored tabs retain their selected view.

- **Tests**
  - Added coverage for view persistence, fallback behavior, tab restoration, and view-switching interactions.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,8 @@
+# Every .js file in this tree is a QML JavaScript library: it opens with the
+# QML-only `.pragma library` directive, which Biome, a plain-JS tool, cannot
+# parse. Its run here can only report syntax errors on files that load fine
+# under qs, so the tool is turned off rather than fed files it cannot read.
+reviews:
+  tools:
+    biome:
+      enabled: false
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,8 +1,0 @@
-# Every .js file in this tree is a QML JavaScript library: it opens with the
-# QML-only `.pragma library` directive, which Biome, a plain-JS tool, cannot
-# parse. Its run here can only report syntax errors on files that load fine
-# under qs, so the tool is turned off rather than fed files it cannot read.
-reviews:
-  tools:
-    biome:
-      enabled: false
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -449,7 +449,9 @@ paths, the lock and the write.
 the merged document and writes nothing. `flea --ui-state '<json object>'` merges that patch through
 the lock and prints the result. The window reaches it from `ui/ViewState.qml` through a `Process`;
 the terminal interface is not here yet, as `flea --tui` says by exiting 2, and when it is built it
-will submit patches for `view`, `hidden` and `sort` only, because menus and places are the window's.
+will submit patches for `view`, `hidden` and `sort` only, because menus and places are the window's;
+`view` is shared between the two, and the window's own patch for it is the startup answer described
+below.
 Scale is on neither list: `src/uischema.rs` has no `scale` key at all, it stores an Omarchy text-size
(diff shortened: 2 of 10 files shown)


thisisgm:

PR #106, opened 2026-09-10
ctrl+click keeps the cursor row in the selection

Yiin:
## What was wrong

Click row A, then Ctrl+click row B: the selection became {B}. Row A dropped out. A plain click leaves the set empty with the cursor on A, and every other path reads that as "A is selected": Ops.targetIndices falls back to the cursor row and Shift+click anchors on it. Ctrl+click went through Pane.toggleSelectAt, which only moved the cursor and toggled B. keys.toml calls the gesture "add the row to the selection", and it did not.

## Fix

Filter.toggleRow, beside extendToRow: with an empty set and the cursor on another row it marks the cursor row first, then moves the cursor to the clicked row, toggles it and anchors there. Ctrl+click on the cursor row itself with an empty set marks it once. Pane.toggleSelectAt calls it. Tap.js and keys.toml are unchanged.

## Verification

- tests/js/filter.js: cursor 2 then toggleRow 4 gives 2,4 with cursor and anchor on 4; a set that already holds the cursor row keeps it once; toggleRow on the cursor row with an empty set gives one mark.
- tests/ui.sh: new case_ctrlclick drives a click then a Ctrl+click in the list, grid and columns views and checks selectedIndices, the cursor and the Shift+click anchor; case_click's expectation moves from "4" to "2,4". It needs omarchy-drive, so it was not run here.
- `./tests/js.sh`, `tools/flea-qmllint-gate`, `git diff --check`: green.

## Summary by CodeRabbit

- **New Features**
  - Added consistent Ctrl+click multi-selection behavior across list, grid, and columns views.
  - Ctrl+click now adds or removes the clicked row without unexpectedly replacing the existing selection.
  - Selection and cursor positioning are handled consistently when starting from an empty selection or clicking an already-selected row.

- **Tests**
  - Added coverage for Ctrl+click selection scenarios, including range extension and repeated selection actions across supported views.


thisisgm:
Merged for v0.3.0, thank you.

PR #108, opened 2026-09-10
Add .rar to the extract-eligible extension list

### LINKED ISSUE #107: .rar missing from the Extract-eligible extension list
\`ui/js/Archive.js\` gates the "Extract" context-menu row on a hardcoded extension list:

\`\`\`js
var EXTENSIONS = [".tar.zst", ".tar.bz2", ".tar.gz", ".tar.xz", ".tgz", ".tar", ".zip", ".7z"]
\`\`\`

\`.rar\` isn't in it, so right-clicking a \`.rar\` file never offers Extract — even though the Rust backend (\`src/backend/archive.rs\`, \`extract_argv\`) already handles it fine: any non-\`.7z\` archive is routed to \`bsdtar\`, which reads RAR archives (read-only, via libarchive) without any special-casing needed.

This list looks like it was derived from "formats Flea can also *create*" (bsdtar's tar variants + zip, plus 7z) rather than "formats Flea can extract." That makes sense for the Compress menu, but RAR is extract-only on Linux by nature (no open-source writer), so it's an odd gap for a file manager — plenty of files people receive are \`.rar\` and they currently have no in-app way to extract them.

**Suggested fix:** add \`.rar\` to \`EXTENSIONS\` in \`ui/js/Archive.js\`. The backend needs no changes since \`extract_argv\` already falls through to bsdtar for any extension other than \`.7z\`.

**Caveat worth noting in the fix/PR:** libarchive's RAR reader is read-only and doesn't cover every RAR variant — plain RAR5 archives and RAR archives using BCJ2/PPMd compression can fail to extract via bsdtar, while older RAR4 archives generally work. Might be worth a friendly error message pointing at \`unrar\` as a fallback when bsdtar chokes on a RAR5 archive, rather than a bare tool error.

Tested on flea 0.2.0.

Auxxed:
## Summary
- `ui/js/Archive.js` gates the "Extract" context-menu row (and the extract destination-name logic) on a hardcoded `EXTENSIONS` array. `.rar` was missing, so right-clicking a `.rar` file never offered Extract — even though the Rust backend already handles it fine: `extract_argv()` in `src/backend/archive.rs` routes any non-`.7z` archive to `bsdtar`, which reads RAR archives via libarchive without any special-casing needed.
- Adds `.rar` to `EXTENSIONS`, fixing both `isArchive()` (menu gating) and `extractDir()` (destination naming), since both derive from the same table.
- Extends `tests/js/archive.js` to cover `.rar` in both functions.

Fixes #107

## Test plan
- [x] `./tests/js.sh` — full suite passes, 2986 checks, 0 failed
- [x] `./tests/js.sh archive` — 21 checks, 0 failed (verified the 2 new assertions actually fail without the fix, by reverting `Archive.js` temporarily and re-running)
- [x] No changes to `src/backend/archive.rs` — extraction already worked via `bsdtar`; this only fixes the UI-side gate

**Known caveat (not addressed by this PR):** libarchive's RAR reader is read-only and doesn't cover every RAR variant — plain RAR5 archives and RAR archives using BCJ2/PPMd compression can still fail to extract via `bsdtar`, while older RAR4 archives generally work. That's a `bsdtar`/libarchive limitation outside this PR's scope; happy to follow up with a friendlier error message pointing at `unrar` as a fallback if that's wanted.

https://claude.ai/code/session_01RUEVNrijEoLhAEenXRx5es

## Summary by CodeRabbit

* **New Features**
  * Added support for recognizing RAR files as archives.
  * RAR files now resolve to their expected extraction directory.


johnkattenhorn:
308d732, "feat: extract rar archives", added the same `.rar` entry to `ui/js/Archive.js` along with `tools/flea-rar-fixture` and coverage in `tests/archive.sh`, and it is in both v0.2.1 and v0.3.0. This branch is based on the line as it stood before that commit, so the change is already on main.

I verified the shipped behaviour on 0.3.0 with no 7zip installed, and libarchive lists and extracts the fixture's rar cleanly.

thisisgm:
Superseded by `308d732` (`v0.2.1`): RAR extraction already shipped independently; this PR was not merged.

Issue #14, opened 2026-09-03
Vulkan-only startup leaves invisible processes on GPUs without Vulkan support


cmyk:

vibefodder:
On QEMU Virtio GPU, Flea 0.1.3-1 crashes at startup because Vulkan initialization fails with VK_ERROR_INCOMPATIBLE_DRIVER (-9), then QuickShell SIGSEGVs in QRhi::create. Setting QSG_RHI_BACKEND=opengl avoids the crash; Flea remains running.

Environment: quickshell-git 0.3.0.r20.g28771c7, Qt 6.11.1, Mesa 26.1.7, Hyprland/Wayland.

cmyk:
Confirmed on a MacBook running Omarchy with Flea 0.1.3-1.

The `SUPER + SHIFT + F` binding reached `uwsm-app` and started `flea --gui`, but Vulkan failed with `VK_ERROR_INCOMPATIBLE_DRIVER`. No window appeared, and repeated attempts left 11 active user scopes, each retaining a windowless `qs` process and its Flea backend. Starting with `QSG_RHI_BACKEND=opengl` opened the window immediately.

I then built and ran v0.1.4 (`dda91be`) on the same machine. Its preflight probe reported:

```
flea: Vulkan is unusable, vkEnumeratePhysicalDevices answered -3, so the shell starts on OpenGL
```

Flea opened successfully on the active workspace without entering the failed Vulkan path. This confirms that `8833fa8` fixes the same failure mode on this hardware too.

As of 2026-09-06, both the Omarchy repository and the AUR `flea` package still advertise 0.1.3-1, so packaged installs do not receive this fix until those repositories publish v0.1.4.

PR #119, opened 2026-09-11
fix: present Vulkan on the GPU that owns the display

gladimdim:
## Problem

On hybrid Intel + NVIDIA Omarchy laptops, Flea opens a mapped window that never paints (it looks like a black/empty file manager).

Hyprland composites on the GPU that owns the connected panel. Qt's Vulkan RHI opens **physical device 0**, which is often the iGPU. The compositor cannot import that buffer.

This was reproduced on an ASUS ROG with Raptor Lake UHD + RTX 4080 Max-Q: Hyprland's primary DRM device is NVIDIA (`DP-1` / `eDP-1` on `card1`), while QRhi logged:

```
Physical device 0: 'Intel(R) Graphics (RPL-S)'    using this physical device
Physical device 1: 'NVIDIA GeForce RTX 4080 Laptop GPU'
```

Forcing `VK_ICD_FILENAMES` to `nvidia_icd.json` draws the listing. Forcing the NVIDIA ICD on **every** NVIDIA laptop would blank classic Optimus, where the compositor sits on Intel.

## Why not `QT_VK_PHYSICAL_DEVICE_INDEX`

`vkEnumeratePhysicalDevices` on this box lists NVIDIA then Intel. QRhi lists Intel then NVIDIA. An index from Flea's probe would point Qt at the other GPU.

## Fix

The existing Vulkan probe now also reads each device's PCI id and matches it against `/sys/class/drm` connectors whose `status` is `connected`.

When Vulkan lists both a display GPU and one that owns no connector, Flea sets `VK_DRIVER_FILES` and `VK_ICD_FILENAMES` to the ICD whose `library_path` belongs to the display GPU, and says so once on stderr.

- Operator-set `VK_DRIVER_FILES` / `VK_ICD_FILENAMES` are left alone (empty is absent).
- A box whose every Vulkan device owns a connected connector is unchanged.
- `intel_hasvk` is skipped when the modern Intel ICD is the display GPU.

## Tests

- Unit tests for DRM connector parsing, ICD JSON `library_path`, vendor matching, and the hybrid vs Optimus vs single-GPU pin rule.
- `tests/modes.sh` asserts the automatic renderer still starts on Vulkan, pins an ICD on a hybrid box, preserves an explicit ICD list, and still falls back to OpenGL when the loader cannot create an instance (issue #14).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Vulkan device selection on hybrid graphics systems to prefer the GPU connected to an active display.
  * Preserved explicitly configured Vulkan driver settings.
  * Avoided overriding default driver selection when all GPUs can provide display output or no display connectors are detected.
  * Improved compatibility with incomplete device enumeration and varied Vulkan or display configurations.

* **Documentation**
  * Updated operator guidance covering Vulkan device selection, configuration overrides, and diagnostic messages.


thisisgm:
Merged in `93e0a090` (`v0.3.2`): Vulkan now presents on the GPU that owns the display, and the notes say the hybrid case rests on your reproduction because minipc has one GPU.

PR #122, opened 2026-09-11
feat(devices): a plugged phone is a rail row that mounts on activation

akitaonrails:
An Android phone plugged in over USB never appeared anywhere in Flea: ui/DeviceMounts.qml enumerates lsblk, and a phone is a gvfs volume with no block device behind it, so only the volume monitors can list one. Nautilus shows it and mounts it on a click; Flea did not. Measured on this box with a Samsung phone (MTP): `gio mount -li` prints the phone's `Volume()` block with its `activation_root` and `can_mount=1`, and the flag is free — 506 ms for `-l` and `-li` alike over three interleaved runs, because the cost is walking the monitors and not the printing.

The phone mounted from the DEVICES rail, browsing Internal storage over the gvfs FUSE path

## What changes

- ui/MountListing.qml: the five second poll asks for `-li`, so the activation root rides the listing three Services already share instead of earning a second 500 ms walk.
- ui/js/Phones.js (new): reads COLUMN-ZERO `Volume()` blocks whose Type line names the two monitors with no block device behind them, `GProxyVolumeMonitorMTP` (Android) and `GProxyVolumeMonitorGPhoto2` (cameras, iPhones), into DEVICES rows. A udisks volume prints indented under its `Drive()` and is the wrong monitor besides, so both guards hold on their own; `can_mount=0` and a missing activation root are not rows anyone can act on. The indented `Mount()` inside the block is what flips a row to mounted.
- ui/PhoneMounts.qml (new): lists and unmounts only. Mounting and opening ride ui/NetworkMounts.qml's existing openShare leg unchanged, because an mtp:// uri mounts, resolves its FUSE folder through `gio info`'s `local path:` line and opens exactly the way a share does — same single-flight guards, same timeouts.
- ui/js/Mounts.js: parseMounts skips `mtp://`/`gphoto2://` the way it already skips `file://` — a mounted phone's shadow `GDaemonMount` (`Mount(1): mtp -> ...`, `is_shadowed=1`) is the volume row's own, so the NETWORK rail never draws the phone a second time under the label "mtp". railMenu offers a mounted phone Unmount (gvfs answers `can_eject=0` for MTP, so it is never Eject), railKey answers its uri, and release routes the phone action through the sidebar, which owns the Service; Ctrl+E therefore works on a phone row for free.
- ui/Sidebar.qml: phone rows sit after the block devices in the DEVICES group, so DeviceMounts's own indices are unmoved; activation dispatches on the row's kind.
- ui/js/Icons.js: a smartphone mark, lucide's rx=2 body rect cut square the way drive's arcs were.
- tests: a new 29-check `phones` suite off the live `-li` capture; tests/js/network.js pins the mtp/gphoto2 skip and keeps the label-not-share-on-host case on a dav fixture; every gio stub that matched `"mount -l"` verbatim (tests/ui.sh × 6, tests/mount-listing.sh, tests/network-open-share.sh) now speaks `-li`.

## What it deliberately does not do (v1)

- No automount: the row appears on plug-in and mounts on activation, because a mount the user never asked for is a mount a locked phone refuses confusingly.
- A failed phone mount reports through the open leg's existing "Connect failed" wording, which reads network-flavored for a phone; a phone-specific sentence would touch the leg this PR deliberately reuses unchanged.
- The label is the volume's own (`SAMSUNG Android`), not the post-mount display name (`Fabio's Z Fold8`): the row's identity should not change under the cursor when the mount lands.

## Verification

- ./tests/js.sh: 3016 checks, 0 failed (29 new); the dst and edmonton batteries green beside it.
- ./tests/mount-listing.sh and ./tests/network-open-share.sh: green with the `-li` stubs.
- tools/flea-qmllint-gate: the three findings it reports are on main before this branch (ColumnsArea unused-imports, TrashMonitor property-override) and none are in the touched files.
- cargo build debug and release: green; no Rust is touched.
- Live on the box, screenshot above: the phone appears under DEVICES while unmounted, a click mounts it and opens Internal storage at its FUSE path, the NETWORK rail stays clean of the shadow mount, Unmount from the row menu releases it, and `gio mount -l` confirms the box back at zero mounts after.
- tests/ui.sh not run here: it drives the pointer through uinput and needs the display session.

https://claude.ai/code/session_01TrpNk5DeqpqyuKRcR6syge

## Summary by CodeRabbit

* **New Features**
  * Added support for detecting and displaying connected USB drives, MTP phones, and PTP cameras in the device rail.
  * Added smartphone icons and device actions for opening and unmounting supported phones and cameras.
  * Improved mount detection for connected mobile devices and cameras.

* **Documentation**
  * Updated the README to document USB, phone, and camera support.

* **Tests**
  * Added coverage for device detection, filtering, mounting, unmounting, and rail interactions.


thisisgm:
Landed in v0.3.0 reworked onto the phone board with credit in the commit body; closing in favour of that commit, thank you.

PR #126, opened 2026-09-12
Show unmounted internal drives in the rail (Fixes #125)

### LINKED ISSUE #125: Unmounted internal drives are hidden from the Devices rail
**Version**: flea-git, `flea --version` reports 0.2.1

**Environment**: Omarchy / Arch Linux (Hyprland), two unmounted internal SATA drives present.

### What I expected

Like most file managers, I expected the Devices rail to list every internal drive (mounted *or* unmounted) so I can click one to mount it from the UI.

### What happens

The two internal drives never appear in the rail because they are neither mounted nor removable:

- `/dev/sdb1` — btrfs, label `backup`, 477 GB, internal (not removable)
- `/dev/sdc1` — ext4, label `x3`, 932 GB, internal (not removable)

`lsblk -o NAME,SIZE,FSTYPE,LABEL,MOUNTPOINTS`:

```
sdb    476.9G
└─sdb1 476.9G btrfs  backup
sdc    931.5G
└─sdc1 931.5G ext4   x3
```

Neither is in `/etc/fstab`, and `udisksctl info` shows them as present but unmounted.

### Root cause

`ui/js/Devices.js` (`collectVolumes`) only emits a volume when it is removable **or** currently mounted:

```js
if (n.name && kids.length === 0 && (n.rm === true || mountOf(n).length > 0))
```

The comment there says: _"An unmounted internal partition stays out: a box's spare EFI and recovery partitions are not places to browse."_

That heuristic makes sense for hiding spare EFI/recovery partitions, but it also hides legitimate extra internal data drives the user wants to mount from the UI.

### Suggested fix

Offer to show unmounted internal volumes too, ideally behind a toggle (e.g. a `flea` setting), since slapping every unmounted partition in the rail would bring back the EFI/recovery noise the current heuristic avoids.

Related: `systemDisk()` already walks the whole tree to find the disk holding `/`, so it is straightforward to extend the walk to carry unmounted internal volumes as click-to-mount rows as well.

taxin-404:
## Summary

Internal drives that are neither mounted nor removable never appear in the Devices rail. `ui/js/Devices.js` only emitted a volume when it was removable or already mounted, so data drives like a spare btrfs `backup` or ext4 `x3` volume were invisible from the UI until something mounted them.

This adds a **Places > Rail** setting, **"Unmounted drives"** (off by default), that also emits unmounted internal volumes as click-to-mount rows (`gio mount -d <device>`). Swap partitions and the system disk stay excluded either way.

## What changed

- **ui/js/Devices.js** — thread `showUnmounted` through the volume walk; keep a `[SWAP]` leaf out even when the toggle is on
- **ui/DeviceMounts.qml** + **ui/Sidebar.qml** — wire the setting through to the rail
- **ui/js/Settings.js** — new "Unmounted drives" rail check row
- **src/uischema.rs** — `showUnmounted` default `false` + `Bool` schema rule so it persists in ui.json

## Testing

- `tests/js/mounts.js` — the toggle adds the unmounted internal partition as a click-to-mount row, and swap stays out even on
- `tests/js/settings.js` — the new row renders and reflects the persisted value
- `src/uistate.rs` — round-trip test: defaults off, rejects a non-Bool, accepts Bool patches
- `cargo test --release`: 605 passed
- pure JS suites: 3096 checks, 0 failed

Fixes #125

## Summary by CodeRabbit

- **New Features**
  - Added an **Unmounted drives** setting under Places.
  - Users can choose to display unmounted non-swap drives in the sidebar.
  - The setting is saved and restored across sessions and is disabled by default.

- **Tests**
  - Added coverage for the new setting, including persistence and swap-drive exclusion.
  - Updated Places and device display checks.


thisisgm:
Landed in v0.3.0 reworked onto the rail board with credit; closing in favour of that commit, thank you.

PR #127, opened 2026-09-12
feat: the chrome bar is the window's title bar for the pointer

PowerSerg10:
## What

The window cannot be moved with the pointer. `ui/shell.qml` asks the compositor for no decorations and draws its chrome itself, so nothing in the tree ever sent an `xdg_toplevel.move`, and the only way to move the window was the compositor's own bind — which a pointer user does not find. Reaching for the top of the window and dragging it did nothing.

`ui/WindowDrag.qml` is that gesture, declared over `ui/ChromeBar.qml`'s strip.

## Why a handler and not an overlay

`DragHandler` with `target: null`, calling `startSystemMove()` once the point has travelled past its threshold — what a title bar does.

A handler and not an `Item` with a `MouseArea`, for the reason `ui/RowDrag.qml` gives: a handler claims the grab only when the point travels, so a press that does not move stays the button, crumb or path under it. An overlay would have had to take those clicks and give them back. The window is reached through Quickshell's attached `Window.window`, the route `ui/CardScroll.qml` and `ui/Ipc.qml` already take, so no window is threaded through the caller's property list and `ui/shell.qml` is not touched at all.

It stands down while the path bar is open (`enabled: !root.editing`), because that editor's own press-and-travel selects the text in it.

Dragging a tiled window pops it out of the layout and re-tiles it on release. That is Hyprland's own rule for a client move request — its `DragController` sets `m_draggingTiled` and `dragEnd` calls `changeFloatingMode` on it — not something this component decides. A floating window drags and stays.

## File budget

`ui/ChromeBar.qml` is at a recorded length in `tools/flea-file-budget` and may not grow, so the two additions there are net zero: the comment on the crumb `TapHandler` drops the rationale that already lives in AGENTS.md under "Testing" and keeps the one line that does not. `ui/shell.qml` is untouched. The gate passes at 448 and 532, both exactly as recorded.

## Verified

Against the running compositor with real `ydotool` input, sampling the window mid-drag:

| Case | Result |
|---|---|
| Drag on the chrome | Window detaches and tracks the pointer; `floating` flips to true |
| Plain click on Settings | Panel opens — the click is not stolen |
| Drag inside the open path editor | Window does not move; text selection preserved |

The path-editor case was confirmed visually, not just by coordinate: the editor was open with `/home/lando/` selected and an accent border.

Local gates: `tools/flea-file-budget` exit 0 with no growth; `tests/js.sh` 3088 checks, 0 failed; `tests/keymap-gen.sh` all ok; `tools/flea-qmllint-gate` introduces no new findings (its two FAILs — `unused-imports: 2`, `property-override: 1` — are pre-existing in `ui/ColumnsArea.qml` and `ui/TrashMonitor.qml`, confirmed identical against a stashed baseline).

## Not verified

**`tests/ui.sh` could not be run here** — it needs `omarchy-drive`, which is not installed on this box. That is the suite most likely to catch a regression in this area, so it should be run before merge. `tests/js/tap.js` asserts the `keys.toml` pointer table length (18) and count-per-place; this change adds no pointer-table row because a drag is not a click and `ui/js/Tap.js` does not decide it, so those counts are unchanged and `tests/js.sh` passes.

## Summary by CodeRabbit

* **New Features**
  * Added title-bar window dragging for undecorated windows using the left mouse button.
  * Dragging begins after a short pointer movement and works with tiled windows.
  * Window dragging is temporarily disabled while editing the path bar.

* **Documentation**
  * Added documentation describing window-drag behavior and interactions with the path bar.


thisisgm:
Landed in v0.3.0 reworked onto the chrome board with credit; closing in favour of that commit, thank you.

PR #129, opened 2026-09-12
Fix file and directory rename on MEGA mounts

nimatrueway:
## Summary

- recognize `fuse.megafs` as a measured FUSE mount that rejects `RENAME_NOREPLACE` with `EINVAL` for files and directories
- reuse the existing exclusive copy-then-remove fallback while preserving `fuse.rclone`'s directory-only scope
- document the compatibility behavior and cover both MEGA source types with regression assertions

## Bug

On a MEGA mount created by `megafs`, Flea 0.2.1 cannot rename files or directories. The backend returns:

```json
{"t":"error","where":"rename","msg":"invalid input"}
```

The mount accepts ordinary `rename(2)`, which is why Nautilus/GIO succeeds, but rejects `renameat2(..., RENAME_NOREPLACE)` with `EINVAL`. Flea previously recognized that response only for directories on `fuse.rclone`.

The fallback retains Flea's no-clobber contract: it creates the destination exclusively, copies the source, and removes the source only after the copy completes.

## Verification

- `cargo test`: 604 passed
- live `fuse.megafs` file and nested-directory renames through the patched backend: success, sources removed, payloads preserved
- live existing-destination attempts: refused with `already exists`, sources and destinations preserved

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved rename compatibility on FUSE-mounted paths.
  - Directory renames on rclone and MEGA mounts now use a compatibility fallback when required.
  - Added support for WebDAV mounts under the user runtime directory.
- **Documentation**
  - Updated protocol documentation to describe all supported compatibility paths and behavior.


src/backend/renamecompat.rs
```diff
@@ -76,13 +76,18 @@ fn needs_gvfs_webdav_fallback(from: &Path, error: &io::Error) -> bool {
...
 
-fn needs_rclone_fallback_in(from: &Path, error: &io::Error, mountinfo: &str) -> bool {
-    error.raw_os_error() == Some(EINVAL)
-        && from
+fn needs_fuse_fallback_in(from: &Path, error: &io::Error, mountinfo: &str) -> bool {
+    if error.raw_os_error() != Some(EINVAL) {
+        return false;
+    }
+    match mount_type_in(from, mountinfo).as_deref() {
+        Some("fuse.megafs") => true,
```
nimatrueway, line 84: should we extend this to all "fuse.*"?

thisisgm:
Merged for v0.3.0, thank you.

PR #130, opened 2026-09-12
feat: show sizes in Miller columns

MaterAnime:
## Summary

- add an optional Size field to the Miller-column view using the existing column settings
- show regular-file sizes immediately and calculate directory sizes asynchronously
- cancel obsolete directory-size work during navigation and sorting
- preserve symlink identity by displaying `link` instead of a misleading byte count
- avoid size metadata reads for path completion, hidden Size fields, and hidden Miller views

Includes the responsiveness fix from #118 by @nerdislb so folder navigation does not wait for obsolete directory-size work.

## Validation

- `cargo test --locked`: 609 passed
- `bash tests/js.sh`: 3,101 checks passed
- `./tools/flea-file-budget`: passed
- `git diff --check`: passed
- release package built successfully with `makepkg`

`tests/protocol.sh` passes the relevant size, cancellation, list, sort, and quit checks. Its three unrelated thumbnail checks still fail because `/tmp/flea-sandbox/flea-media-btrfs/photo_0.jpg` is absent. `flea-qmllint-gate` also retains the existing unused-import, property-override, and incompatible-type baseline findings.

## Review notes

The Size field can be disabled from the existing column settings. When disabled, neighbouring-column peeks do not stat files merely to calculate hidden values.


thisisgm:
Shipped in v0.3.0 as 9d7c3799: the active column draws sizes through the same viewport-scoped folder size path, reworked from this PR and credited in the v0.3.0 notes, thanks @MaterAnime.

Issue #110, opened 2026-09-11
Recursive copy can follow a replaced directory outside the selected tree


gardnmi:
Tested on Linux x86_64 against Flea v0.2.0, commit `b992e76475731041b0c13b50f5cbec3e5c04faa4`. Findings use dummy data only. No installation, real credentials, live shares, or user documents were used.

A search of public open and closed issues and PRs, including targeted searches across comments, found no matching report at review time (2026-09-10). This does not rule out private reports or unindexed discussions.

## Impact and prerequisites

Another actor who can rename or replace a directory in the source tree while Flea is copying it can redirect subsequent reads through a symlink. Flea then reads files outside the selected tree using the victim user's permissions and writes those bytes into the destination.

Actual disclosure requires the attacker to be able to change the source pathname, arrange suitable child names, and read the destination (or otherwise receive the copied data). A shared local directory is a relevant threat model. This is not an attack triggered just by viewing any ordinary file, and it does not allow reading files the victim account cannot access.

The fixture deterministically models the directory change through the existing progress callback. It imports unchanged production modules and does not alter the copy implementation. It does not demonstrate a separate-UID attacker, network filesystem behavior, or UI timing.

## Reproduction

With Rust available, run:

```bash
python3 reproduce-directory-swap.py /absolute/path/to/flea
```

The script compiles a small harness that imports the repository's copy modules. It creates two harmless selected files and a separate dummy private directory containing same-named files. After the first copy progress callback, it renames the selected source directory aside and puts a symlink to the outside directory at the original source path. Whichever file is enumerated second is then read through the replacement directory.

Observed output (which filename appears first may vary):

```text
copy result: Ok(())
a.txt: PRIVATE DUMMY CONTENT OUTSIDE SELECTED TREE
b.txt: PUBLIC DUMMY CONTENT
outside-selection files copied: 1
```

Expected: continue reading the originally opened directory through held descriptors, or refuse the operation when the source identity changes. Never import data from the replacement symlink target.

## Cause

[copy_dir_entries, copyfile.rs:118](https://github.com/thisisgm/flea/blob/b992e76475731041b0c13b50f5cbec3e5c04faa4/src/backend/copyfile.rs#L118) obtains directory entries and then passes `entry.path()` back into pathname-based recursion. The directory iterator refers to the old directory, but each child path is resolved again through the now-replaced parent. The `O_NOFOLLOW` on the regular-file open protects only the final path component, not intermediate directories.

[...]

PR #135, opened 2026-09-13
fix: explicitly release flock lock before closing fd

reverb256:
## Problem

On Linux, release of a flock lock when a file descriptor is closed is not sequenced after all observable effects of close(2). Under high thread contention, another thread can observe the lock as still held even after File::drop has run.

This caused intermittent test failures in:
- recovery_cleans_a_completed_deletion_and_its_record
- recovery_cleans_restore_metadata_only_for_its_verified_destination

## Fix

1. Added a Drop impl for Manifest that calls flock(fd, LOCK_UN) explicitly before the File is closed, but only when the Arc reference is the last strong reference.
2. Simplified open_inactive to open directly by path with O_RDWR | O_NOFOLLOW instead of the open_if_regular + /proc/self/fd double-open approach.

## Verification

All 604 tests pass with --test-threads=32 (20 consecutive runs) and --test-threads=64 (10 consecutive runs).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when accessing inactive trash manifests.
  - Ensured file locks are released correctly after the final use of a manifest.
  - Reduced the risk of access conflicts during trash-management operations.
  - Improved compatibility across supported Linux environments.


reverb256:
Addressed the MSRV flag: replaced all three std `File::lock`/`try_lock`/`unlock` call sites in this file with direct `flock(2)` syscalls (identical semantics, no new dependency), so the declared Rust 1.77 MSRV holds for this file. Verified: 605 passed, 0 failed with `--test-threads=32`.

Two notes for the maintainer:
1. The 1.89 requirement predates this PR — `main` already used `File::lock`/`try_lock` here, so the MSRV declaration was already stale.
2. `src/uistore.rs` (`take_lock`) still uses `File::lock` and will need the same treatment or an MSRV bump to 1.89 — left out of scope for this PR.

thisisgm:
Landed in v0.3.0: the flock release half with credit; the by-path reopen half is declined because it reintroduces the #110 inode race. Closing, thank you.

PR #137, opened 2026-09-13
fix: open the folder menu from Places

### LINKED ISSUE #136: Places folders do not open a context menu
Right-clicking Home or a folder such as Downloads under Places does nothing. Trash opens a context menu, but the folder rows do not. This also prevents adding those folders to Favorites directly from Places.

To reproduce on current main (`c6a0149`):

1. Open any directory in Flea.
2. Right-click a folder under Places.
3. No menu appears.

Expected: the normal folder context menu, including copy, rename, trash and Add to Favorites. Actions should target the clicked Places folder even when another directory or selection is open in the listing.

Fix prepared and tested in https://github.com/thisisgm/flea/pull/137.

AksharP5:
Right-clicking folders under Places currently does nothing. These rows now open the shared folder menu, and each action targets the clicked folder independently of the current listing. The sidebar menu key uses the same path.

The existing menu snapshot captures the folder's identity before operations run. Rename uses the existing dialog and follows the new path when the renamed folder contains the current location.

Fixes #136.

Validation:

- Full JavaScript suite and 9 menu backend tests in debug and release passed; file budgets and shell loading passed.
- New backend regressions cover copy/rename targeting, replaced-folder refusal and missing-target refusal. All 8 added assertions passed.
- Isolated real-shell checks passed for the sidebar signal, full menu and submenu, copy, Properties, rename failure/retry, ancestor rename and returning to listing actions.
- `tests/run-all.sh` has five failing suites here: `ops`, `protocol` and `uistate` also fail on unchanged main; `thumbs` and `media` require missing media fixtures. Native pointer tests were unavailable because `omarchy-drive` is not installed.

## Summary by CodeRabbit

- **New Features**
  - Added context menus for Home and Places folders.
  - Folder menus support opening, terminal access, permissions, copying, duplicating, renaming, trashing, pasting, compressing, and adding favorites.
  - Actions apply to the clicked folder, even when another directory is open.
  - Added folder-specific rename handling with confirmation and safeguards against acting on changed or missing folders.

- **Documentation**
  - Documented the new Places-folder menu behavior and supported actions.


thisisgm:
Landed in v0.3.0 reworked onto the menu board with credit; closing in favour of that commit, thank you.

PR #145, opened 2026-09-13
feat: larger grid thumbnails, stepped separately from text size

calebhat:
## Why

Settings → Preview → Thumbnail size tops out at Extra large (128 px). The backend already writes Freedesktop `large` thumbs at 256 px (`THUMB_SIZE` in `src/backend/thumbs.rs`), so the grid was only drawing half of that. Ctrl+Shift+/- is Issue 9 text size, not tile size, so it is the wrong lever for this.

## What

- Keep named stops. Extra large stays **128** so existing `ui.json` does not jump.
- Add **XX-large (192)** and **Huge (256)**, matching the cache the backend already fills.
- One table in `ui/js/ThumbSize.js` for settings, Ctrl+scroll, and the new chords.
- **Ctrl+Alt+= / +** up, **Ctrl+Alt+-** down, **Ctrl+Alt+0** reset to medium. Text zoom is unchanged.
- Status bar announces the pixel size when a chord fires.

## Test

- `tests/js.sh thumbsize settings keymap`
- `tests/keymap-gen.sh`
- `tests/budget.sh`
- `cargo test uischema`
- `tests/acceptance-matrix.sh`

Not run: `tests/ui.sh` (needs the display).

## Summary by CodeRabbit

* **New Features**
  * Added six preview thumbnail sizes, including XX-large and Huge.
  * Added keyboard shortcuts to increase, decrease, or reset thumbnail size.
  * Added Ctrl+scroll support for resizing thumbnails.
  * Thumbnail size changes now display pixel values and status announcements.
* **Documentation**
  * Updated the keymap help sheet with thumbnail-size shortcuts.


calebhat:
Superseded by #147, which includes this size-stops commit plus the scroll/cache work.

PR #149, opened 2026-09-13
feat(picker): show hidden files with the toggleHidden key

### LINKED ISSUE #148: Portal file chooser has no way to show hidden files
**What happens**

The portal file chooser — the Open/Save dialog xdg-desktop-portal routes to Flea through
`portals.conf` — never lists dotfiles, and nothing in that window can reveal them. The listing
request is sent with `hidden: false` unconditionally, the chooser binds no `toggleHidden` key and
draws no context menu, so there is no key, click or menu row that turns them on. The main Flea
window can do this today (`.` and "Show hidden files" in the listing/background menu), so the two
surfaces disagree.

**Impact**

Any file dialog a third-party app raises cannot reach a hidden file or a dot-directory — the common
cases being `~/.config`, `~/.local`, `~/.ssh`, `~/.gnupg`. The chooser has no location field either,
so a hidden directory cannot even be typed in. The only workaround is to stop using Flea as the
FileChooser backend while you need one.

**Steps to reproduce**

1. Have Flea claim the chooser (`flea --picker`), with
   `org.freedesktop.impl.portal.FileChooser=flea;gtk` in `~/.config/xdg-desktop-portal/portals.conf`.
2. From any app that raises a portal file dialog (a browser upload, an editor's Open), browse to your
   home directory.
3. Dotfiles are absent. Pressing `.` — the window's own shortcut — does nothing, and there is no
   menu to look in.

**Expected**

Some way to reveal and hide dotfiles in the chooser, consistent with the window's own toggle: at
minimum the same key, ideally a visible entry as well.

**Environment**

- Flea 0.2.1 (`flea 0.2.1-3`, `omarchy pkg add flea`)
- xdg-desktop-portal 1.22.1, xdg-desktop-portal-hyprland 1.4.1
- Hyprland 0.56.2

felixzsh:
Fixes #148 

The chooser sent `hidden: false` on every listing and handled no `toggleHidden` key, so a dialog a
caller raised could never show or reach a dotfile. `.` — or the preset's `toggleHidden` chord — now
flips the flag on `ui/PickerList.qml` and re-reads the standing directory, the same re-read
`ui/Pane.qml` makes; `ui/picker.qml` builds its list request with that flag. No backend, protocol or
keymap change: both already exist.

`ui/picker.qml` is a recorded over-cap file the ledger may not grow, so it stays at 635 lines;
`tools/flea-file-budget` exits 0 and `tools/flea-qmllint-gate` shows the same pre-existing findings as
`main`. Driven by hand through `flea --pick`: dotfiles appear on `.` and hide on a second press.
`tests/picker.sh` needs `omarchy-drive` and was not run.

## Summary by CodeRabbit

* **New Features**
  * Added a keyboard action to show or hide dotfiles in the file picker.
  * The current directory refreshes automatically when hidden-file visibility is toggled.

* **Documentation**
  * Updated installation guidance to explain the dotfile toggle behavior and related keyboard shortcuts.


thisisgm:
Merged for v0.3.0, thank you.

PR #166, opened 2026-09-17
Proposal: Nautilus-style native rename fallback for unsupported no-replace flags

nimatrueway:
## Motivation
This proposal comes from frustration with the performance of MEGA renames. On a MEGA FUSE mount that rejects `RENAME_NOREPLACE`, renaming a file or directory can fall into a copy/remove path, potentially downloading and uploading contents for what should be a metadata operation. A simple directory rename should not need to transfer the whole tree when the mount supports native rename.

Rather than add another MEGA-specific exception, this draft proposes moving toward the native-move-first behavior of a more mature file manager such as Nautilus, while retaining Flea’s atomic no-overwrite operation wherever it works. The desired collision policy is refusal, not an overwrite or merge prompt.

## Proposed Behavior
1. Try `renameat2(RENAME_NOREPLACE)` first.
2. On `ENOSYS`, `EOPNOTSUPP`, or `EINVAL`, check the destination with `lstat`.
3. Refuse an existing entry, including a dangling symlink. Refuse lookup errors other than not-found.
4. If the destination is absent, attempt ordinary native rename.
5. Copy/remove only on `EXDEV` in the existing move/rename callers. Other errors remain failures.

`EINVAL` is not proof that flags are unsupported: it can also describe an invalid operation. It permits one native retry, whose own result decides success or failure; it does not by itself authorize a recursive copy.

This removes the filesystem-name checks and their now-unused mountinfo parser. The shared helper covers renames, moves, undo, and archive publication. The protocol documentation describes the changed guarantee.

## Nautilus Comparison
In the inspected Nautilus 50.3.1 and GLib 2.88.3 sources, local-path operations use checked ordinary rename/move without selecting a cloud-provider name. Mounted FUSE paths can use that same local backend. Nautilus attempts backend moves with `G_FILE_COPY_NO_FALLBACK_FOR_MOVE` first and handles the items requiring fallback separately.

This is an alignment proposal, not a claim of exact parity. Flea still attempts atomic no-replace first, refuses conflicts instead of presenting replacement/merge choices, and rejects non-ENOENT destination lookup errors. GLib’s local move implementation is less strict about those lookup failures. Remote URI backends can also differ from local mounted paths.

Sources:
- https://github.com/GNOME/nautilus/blob/50.3.1/src/nautilus-file-operations.c (`move_file_prepare`)
- https://github.com/GNOME/glib/blob/2.88.3/gio/glocalfile.c (`g_local_file_move`, `g_local_file_set_display_name`)

## Tradeoffs for Review
- **The fallback is not atomic no-clobber.** Another writer can create the destination after the check and ordinary rename can overwrite it. The inspected GLib local checked-rename path has this race too. The atomic first attempt retains the stronger guarantee on filesystems supporting it.
- The old rclone EINVAL copy exception is replaced by the generic native retry. A failed native rename no longer starts a copy unless it reports EXDEV.
- The old WebDAV EIO copy exception is removed. EIO now fails rather than being treated as an unsupported-capability signal.
- Archive publication also uses this helper and therefore inherits the weaker collision guarantee on the fallback path.
- Native calls can still block on a stalled mount. This does not add asynchronous I/O or deadlines.
- Historical provider-specific descriptions in AGENTS.md remain to be reconciled if this policy is accepted; the protocol document identifies the superseded policy.

## Validation
- 703 Rust tests passed in both debug and release profiles, with no compiler warnings.
- Four focused checked-rename tests cover all three eligible errors, inode-preserving moves and reversals of files/directories/symlinks, collision refusal, unrelated errno passthrough, and lookup/native-operation failures.
- Ten rename-compatibility tests cover EXDEV-only copying, collision refusal, cleanup, and preservation of the complete copy after partial source removal.
- Package checks passed: 3,851 JS checks, the additional QML/keymap probes, and generated-keymap consistency checks.
- Editor diagnostics, whitespace checks, and the file-budget gate passed (existing budget warnings remain).

The fallback tests inject the initial unsupported-flag errors and then use real sandbox filesystem operations. They are not live MEGA/rclone/WebDAV validation or a performance benchmark. No live MEGA data was touched. The package was built for local testing, but installation was blocked by sudo authentication, so installed-product testing is still pending.

Keeping this as a draft to discuss whether this compatibility/performance tradeoff is acceptable before treating it as a replacement for the current no-clobber policy.

## Prior Proposal
For context, my earlier MEGA-specific rename PR is [#129: Fix file and directory rename on MEGA mounts](https://github.com/thisisgm/flea/pull/129). This draft revisits that problem with a provider-independent checked-native-rename strategy motivated by the performance frustration described above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Rename operations now consistently refuse existing destinations, including dangling symbolic links.
  * Cross-device moves continue to use copy-and-remove, while other rename errors now fail without starting a recursive copy.
  * Rename behavior is more consistent across storage providers and filesystem types.
  * Files, directories, and symbolic links retain their identity when renamed successfully.

* **Documentation**
  * Updated protocol documentation to reflect checked native rename behavior and cross-device move handling.


thisisgm:
Declined: checking then renaming loses atomic no-clobber safety; #170 remains unreproduced on `643313174f3778d561c95abeafe845dcdaeab11c` (`v0.3.1`) and stays open.

nimatrueway:
> Declined: checking then renaming loses atomic no-clobber safety; #170 remains unreproduced on `643313174f3778d561c95abeafe845dcdaeab11c` (`v0.3.1`) and stays open.

I suspect that's what users prefer over copy-then-remove fallback; having to wait several seconds or minutes for a simple rename/move is no-bueno user experience. Perhaps why Nautilus opted for non-atomic move as lesser evil.

thisisgmultra:
I'll have a look at alternatives, don't worry this will not be ignored

PR #175, opened 2026-09-19
Keep file chooser responsive during stalled filesystem reads

nerdislb:
## Summary
- Keep chooser navigation responsive when a read-only listing backend stalls on FUSE/network I/O.
- Replace and reap only chooser-owned listing workers; discard obsolete replies and keep selection validation separate.
- Persist the portal reply before shutting down both read-only children, including failed-start/cancellation races. Normal file-manager write draining is unchanged.

## Verification
- Six real Quickshell process lifecycle cases, including blocked listing/validation, stale output, cancellation and failed startup; controlled signal-order regression fails with the previous guard.
- 68 picker JavaScript checks pass.
- Actual release binary + complete picker UI selects and accepts a local fixture.
- Native desktop acceptance: Space/Enter selection, Backspace navigation away from a controlled stalled reader, Escape cancellation, and window-manager close all produce the expected portal reply and reap owned helpers.
- Small warm local navigation sample: replacement median 3 ms versus reused-worker median 0.5 ms (10 samples each, 1 ms clock resolution); not a general performance benchmark.
- Independent Claude Fable review; startup race corrected and re-reviewed.

## Scope / baseline
Independent of #172 and based on current main. No production mount restart or configuration changes required. SIGKILL cannot guarantee completion for genuinely uninterruptible kernel I/O.

Release build passes. The file-budget check passes on the actual main baseline. No Rust source changes. The existing signal-arity gate fails on both base and candidate after the prior Messages.js extraction; no full-suite-green claim.

The local tracking ref initially lagged because this checkout fetched only a tag. This branch now explicitly includes actual main (`f738261`); the PR diff remains limited to ten picker files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved file chooser reliability when directory listings stall or return incomplete data.
  - Newer navigation requests now supersede outdated requests, helping the chooser recover without displaying stale results.
  - Improved cancellation and shutdown handling, including failed or interrupted background operations.
  - Prevented file-manager write activity during read-only chooser listing operations.
- **Tests**
  - Added coverage for stalled listings, navigation changes, cancellation, missing resources, and shutdown ordering.


thisisgm:
Merged in `aba90892` (`v0.3.2`): the chooser stays responsive through a stalled listing, and Cancel completes even when its backend never starts.

PR #177, opened 2026-09-19
fix: pin a UTF-8 locale inside the archive jail

### LINKED ISSUE #156: Archives with non-ASCII names never extract: the bwrap jail runs bsdtar with --clearenv and no locale
## Summary

Extracting an archive whose members have non-ASCII names always fails, and nothing is written to the destination. The extract jail in `src/backend/sandbox.rs` passes `--clearenv` to bwrap and re-sets only `MALLOC_ARENA_MAX`, so `bsdtar` runs with no `LANG` and no `LC_ALL`. glibc then resolves the locale to `C`, and libarchive refuses to convert the archive's UTF-8 pathnames into that charset, skipping every member:

```
bsdtar: Pathname cannot be converted from UTF-8 to current locale
bsdtar: Archive entry has empty or unreadable filename ... skipping
bsdtar: Error exit delayed from previous errors
```

The user sees the last line and an extract that produced nothing. The host locale is irrelevant — mine is `en_US.UTF-8` and the same archive extracts fine from a shell — because `--clearenv` drops it before `bsdtar` starts.

This is the same family as #36 (a child process whose locale is not pinned). There the problem was translated output being parsed as English; here it is the charset.

## Reproduction

flea 0.2.1-3, Omarchy 4.0.4-1, libarchive 3.8.9, bubblewrap 0.12.0, kernel 7.1.9-arch1-2, host `LANG=en_US.UTF-8`, with `C.utf8` and `en_US.utf8` generated.

Any archive with an accented member reproduces it. A Notion export whose top-level entries are `Remoção de Viés 1º Mês/` and `Remoção de Viés 1º Mês <id>.md` is what found it here.

In Flea: select the `.zip` → Extract → error, no destination directory.

From a shell, with the host locale, the same archive is fine:

```
$ bsdtar -xf export.zip ; echo exit=$?
exit=0          # 15 entries, names intact
```

Replaying `BWRAP_FLAGS` exactly as `sandbox.rs` builds them today:

```
$ FLAGS=(--unshare-all --die-with-parent --new-session --clearenv \
         --setenv MALLOC_ARENA_MAX 2 \
         --ro-bind /usr /usr --ro-bind /etc /etc \
         --symlink usr/lib /lib --symlink usr/lib /lib64 \
         --symlink usr/bin /bin --symlink usr/bin /sbin \
         --proc /proc --dev /dev --tmpfs /tmp)

$ bwrap "${FLAGS[@]}" --ro-bind "$PWD/export.zip" /in.zip --bind "$PWD/out" /out \
        /usr/bin/bsdtar -x -f /in.zip -C /out
bsdtar: Pathname cannot be converted from UTF-8 to current locale
bsdtar: Archive entry has empty or unreadable filename ... skipping
bsdtar: Error exit delayed from previous errors
$ ls -A out | wc -l
0
```

The same command with a UTF-8 locale added to the jail:

```
$ bwrap "${FLAGS[@]}" --setenv LC_ALL C.UTF-8 \
        --ro-bind "$PWD/export.zip" /in.zip --bind "$PWD/out" /out \
        /usr/bin/bsdtar -x -f /in.zip -C /out
$ ls out
'Remoção de Viés 1º Mês'  'Remoção de Viés 1º Mês 2273a90c01054c0085413457ad404217.md'
```

## Scope

Measured on this box, under the flags above:

- **Extract** — broken. Every member is skipped, `bsdtar` exits 1, the staged directory stays empty and nothing is published.
- **Listing** (`bsdtar -tvf`, the archive index) — broken the same way: every row is skipped and the command exits 1.
- **Compress** — unaffected. Names go in as bytes, and a `.zip` built inside the jail from a file named `Relatório Março.txt` reads back correctly.
- **ASCII-only archives** — unaffected, which is presumably why the test corpus does not catch this.

## Suggested fix

Add one pair to `BWRAP_FLAGS` in `src/backend/sandbox.rs`:

```rust
"--setenv",
"LC_ALL",
"C.UTF-8",
```

`C.UTF-8` rather than passing the host's `LANG` through, on purpose: it gives libarchive a UTF-8 charset while keeping tool messages and the listing columns in English. Inheriting a translated host locale would fix the encoding and reintroduce #36's problem one layer down, since `archivespec.rs` reads `bsdtar -tvf` by column position.

`/usr` is already `--ro-bind`ed into the jail, so `/usr/lib/locale/locale-archive` is present and no extra bind is needed — only the variable is missing.

Worth checking whether the other jailed tools (thumbnailers, `convert`) have a latent version of this on non-ASCII paths.

Happy to test a patch on this box.

johnkattenhorn:
`--clearenv` in `BWRAP_FLAGS` leaves the archive jail in the POSIX locale. bsdtar will not convert a UTF-8 member name into it, so it exits 1 and `run_boxed` reports the extract as failed. This sets `LC_ALL=C.UTF-8` inside the jail.

Closes #156.

## Measured

Flea 0.3.0, libarchive's bsdtar. A tar.gz holding `café.txt`, `документ.txt` and `日本語.txt`, run through the exact flags `wrap()` builds:

```
bsdtar: Pathname can't be converted from UTF-8 to current locale
bsdtar: Pathname can't be converted from UTF-8 to current locale
bsdtar: Pathname can't be converted from UTF-8 to current locale
bsdtar: Error exit delayed from previous errors
exit=1
```

Same jail with `--setenv LC_ALL C.UTF-8` added: `exit=0`, all three names correct on disk.

Two details the ticket does not record:

- The files land with correct bytes in both runs. Only the status is wrong. `run_boxed` (`archivework.rs:78`) fails on a non-zero exit and surfaces the last stderr line, so the operator reads `Error exit delayed from previous errors`, which names neither the locale nor a file.
- `wrap_readonly` needs it too. `bsdtar -tvf` in the same jail prints `\346\227\245\346\234\254\350\252\236.txt`, so the archive index in the preview pane is mangled for an operator who never extracts anything.

## Why C.UTF-8

glibc builds it in. `locale -a` inside the jail, with only `/usr` and `/etc` bound, lists `C.utf8` and `POSIX`, with nothing generated and nothing extra bound. The jail cannot inherit the caller's `LANG` through `--clearenv`, and should not.

The tree already pins a locale where it runs a tool. `menu_registry.rs:43` and `trashbrowse.rs:86` both set `LC_ALL=C`, and 49d6cd0 is a test written to record why. The bwrap jail does not, which is what this changes.

## Tests

- `sandbox.rs`: `both_jails_pin_a_utf8_locale`, static, asserts both wrappers carry the flag.
- `archiveops.rs`: `an_archive_whose_members_are_not_ascii_extracts`, live, behind `sandboxprobe::skipped()`.

Deleting the three flag lines and keeping the test fails it with the message from the report:

```
panicked at src/backend/archiveops.rs:353:44:
a UTF-8 member name is not an extract failure: FleaError { where_: "archive", path: "", msg: "bsdtar: Error exit delayed from previous errors" }
```

The test uses three scripts rather than one because a Latin-1-representable name on its own would pass against a jail that had fallen back to an eight-bit non-UTF-8 locale.

`cargo test --locked`: 706 passed, 0 failed. `tests/js.sh` and `tests/keymap-gen.sh` green. Debug and release both build with zero warnings. `tools/flea-file-budget` exits 0; `sandbox.rs` at 283 and `archiveops.rs` at 366 are over the 250 soft budget and under the 400 hard cap.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed archive extraction and listing for archives containing non-ASCII filenames.
  * Ensured UTF-8 filenames from multiple writing systems are handled successfully during extraction.


thisisgm:
Superseded by `5094f868` (`v0.3.1`): the same UTF-8 sandbox fix shipped independently; this PR was not merged.

PR #184, opened 2026-09-21
test(shelfundo): keep the stranger's inode distinct from the move's

### LINKED ISSUE #180: shelfundo: undo_refuses_to_walk_a_stranger_back fails when the filesystem reuses the inode
`shelfundo::tests::undo_refuses_to_walk_a_stranger_back` fails on filesystems that
hand the just-freed inode to the next file. It fails the Omarchy package build
`check()` and keeps 0.3.0/0.3.1 out of the repo.

Observed on the Omarchy builder, building the v0.3.0 tree:

```
thread 'shelfundo::tests::undo_refuses_to_walk_a_stranger_back' panicked at
src/shelfundo_tests.rs:67:5:
/tmp/flea-test-shelfundo-stranger-.../was/a.txt could not go back (file or folder not found)
test result: FAILED. 627 passed; 1 failed; 0 ignored; 0 measured; 76 filtered out
```

The test writes a file, records its identity, removes it, writes a different file
at the same name, and expects `put_back` to refuse with "was replaced since the
move". `put_back` (`src/shelfundo.rs:215`) compares only dev, ino and kind:

```rust
if meta.dev() != step.dev || meta.ino() != step.ino || meta.mode() & 0o170000 != step.kind {
    return Err(format!("{} was replaced since the move, so undo left it in place", step.to));
}
```

and the shelf `Move` stores exactly those three, because `ItemIdentity::parts()`
drops the ctime that `ItemIdentity::record()` captures (`src/backend/undo.rs:16-27`).

When the recreated file lands on the same inode the check reads "same item", so it
does not refuse; it falls through to `move_back(to -> was/a.txt)`, whose parent does
not exist, and returns the wrong error. That is the observed failure.

This looks like a real hole and not only a test artifact: dev+ino+kind cannot tell
a recreated file apart when the inode is reused, so shelf undo can move the wrong
file back. The identity already records ctime; the shelf move just does not carry it.

Reproduces under tmpfs/ext4, where the allocator reuses the freed inode immediately.

Possible directions:

- carry ctime (and nsec) into the shelf `Move` identity and compare it in `put_back`; or
- force a distinct inode in the test; or both.

Happy to open a PR if you prefer one direction.

Context: Omarchy's package `check()` runs this suite and aborts the release on it
(omarchy-pkgs#469, #411). v0.3.1 and `main` still carry the same test and the same
`put_back`, so the version bump does not fix it.

felixzsh:
`shelfundo::tests::undo_refuses_to_walk_a_stranger_back` removes the landed file, writes
a replacement at the same name, and expects the recorded `dev`/`ino`/`kind` to refuse it.
On a filesystem that hands the just-freed inode to the replacement -- ext4 does, and the
Omarchy package builder's `/tmp` did -- the stranger carries the move's own identity,
`put_back` does not refuse, and it falls through to `move_back` with the source directory
absent, failing with "could not go back (file or folder not found)".

That one failure aborted `check()` and kept flea 0.3.0/0.3.1 from publishing. Omarchy
worked around it in the package by running the `shelfundo` suite with `TMPDIR` on
`/dev/shm` (tmpfs, which allocates inode numbers from a counter and never reuses one),
which is exactly the allocator behaviour this test was silently assuming. This makes the
test independent of that assumption, so it holds on ext4, tmpfs and anywhere else.

Hold the landed file open across the unlink, so its inode stays allocated and the
replacement is provably a different item. `dev`+`ino`+`kind` remains the identity: `ctime`
is deliberately not part of it because permission repair changes it (see
`backend/opsreq.rs`), so the test has to make the two items differ rather than trust the
allocator.

Verified: the `shelfundo` suite is 8 passed on btrfs and tmpfs; the original failure does
not reproduce here (this box's filesystems do not recycle the inode), which is why it only
showed on the builder.

Fixes #180.

## Summary by CodeRabbit

- **Tests**
  - Improved reliability of undo behavior tests when a destination file is deleted and recreated.
  - Ensured the test distinguishes the recreated file from the original file by keeping the destination open during deletion.


thisisgm:
Shipped in v0.3.3. Thanks @felixzsh! https://github.com/thisisgm/flea/releases/tag/v0.3.3

